PR #154, opened 2021-08-05
Migrate to Next

mattpocock:
- [x] Find a solution for react-json-view
- [x] Fix editor theme
- [x] Change base path to /viz
- [x] Fix Cypress tests

### Limitations

1. I've not yet built out SSR for the pages, so the OG Image tags still work as normal. This can be brought in later as a separate PR.

2. Everything is rendered on client-side to avoid Monaco weirdness:

https://github.com/statelyai/xstate-viz/pull/154/files#diff-88a36bc7a53bfa58c6f451464798d631d785160dfacfdaf9479be7a0b1d7e5e5R33-R48

I think this is a sensible way to carry on for now. There's no huge benefit to us to actually SSR the HTML, because we have so much computation to do in the client-side anyway (loading Monaco, ELK, prettier) that pre-rendering the HTML is a drop in the ocean.

### Benefits

Local development is now much closer to how the app will actually be deployed - the viz is at `/viz` and has a local rewrite to the registry at `/registry`, meaning you can test the integration between them fully.


src/EditorWithXStateImports.tsx
```diff
@@ -130,14 +130,13 @@ export const EditorWithXStateImports = (
           onMount={async (editor, monaco) => {
             editorRef.current = monaco.editor;
             const theme = editorTheme.theme;
-            monaco.editor.defineTheme(theme, themes[theme]);
-            monaco.editor.setTheme(theme);
+            editorRef.current.defineTheme(theme, themes[theme]);
+            editorRef.current.setTheme(theme);
```
mattpocock, line 134:
```suggestion
            monaco.editor.defineTheme(theme, themes[theme]);
            monaco.editor.setTheme(theme);
```

src/pages/_app.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <link rel="icon" href="/favicon.png" />
+        <title>XState Viz</title>
+        <meta
+          name="description"
+          content="Visualizer for XState state machines and statecharts"
+        />
+        <script src="https://unpkg.com/prettier@2.3.2/standalone.js"></script>
+        <script src="https://unpkg.com/prettier@2.3.2/parser-typescript.js"></script>
+        <script src="https://unpkg.com/elkjs@0.7.1/lib/elk.bundled.js"></script>
+      </Head>{' '}
```
mattpocock, line 47:
```suggestion
      </Head>
```

tsconfig.json
```diff
@@ -18,10 +18,13 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx",
+    "jsx": "preserve",
```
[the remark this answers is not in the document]
mattpocock: Hm, it's setting it automatically. What's the benefit of keeping react-jsx?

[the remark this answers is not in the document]
Andarist: Disregard that - it doesn't matter in our situation.

cypress.json
```diff
@@ -2,5 +2,5 @@
   "baseUrl": "http://localhost:3000",
   "viewportWidth": 1280,
   "viewportHeight": 800,
-  "defaultCommandTimeout": 8000
+  "defaultCommandTimeout": 15000
```
Andarist, line 5: 😅

Andarist (approved):

farskid (approved):

PR #182, opened 2021-08-13
Only handle platform-specific meta/ctrl keys in `onWheel` responsible for zooming canvas

Andarist:
I'm using a special accessibility setting on Mac that allows me to zoom screen when wheeling with ctrl key being pressed:
Screenshot 2021-08-13 at 13 32 29

Since we are using `e.ctrlKey || e.metaKey`, regardless of the platform, to determine if the canvas should be zoomed when wheeling, I've noticed a weirdo effect on my machine - I zoom both the screen and the canvas, at the same time:
https://www.loom.com/share/5b866dbf55ea4dd5a3fad013348df61c

Note that this accessibility option is configurable and one could bind it to the meta key but I think it's less common since ctrl is the default. We could just `ev.preventDefault()` here but `onWheel` is a passive listener by default in React so I would have to move this to an effect and prevent it there. I'm a little bit worried about perf implications because, well, passive listeners exist for a reason - so I didn't want to go with that solution for the time being. I think this wouldn't break accessibility because the action that we are performing is very similar to what the user tries to do anyway - it's all about zooming.

I think it makes sense to make the key platform-dependent here though, regardless if we'd like to prevent the default in the future or not.


santicros:
Tried this without the preventDefault and this is the behaviour on Firefox Mac with trackpad + CMD. It zooms the canvas and then the browser zooms the page:
https://user-images.githubusercontent.com/24988424/129408647-8131746e-bdde-4489-b5ef-fabe9233c6e4.mov

mattpocock:
Pulled dev into here in order to allow Vercel to preview deploy it. Want to test on my Windows machine

mattpocock (changes requested):
Zooms the whole page on Windows. Tested on Chrome

Andarist:
@mattpocock could u retest that with a PR that has been branched off this one: https://github.com/statelyai/xstate-viz/pull/186 ? They both will go into `dev` one after another - I just would prefer to not merge them together to keep the merging history cleaner and changesets behavior more predictable

mattpocock:
@Andarist As mentioned on that PR, I can't test it because the preview deploys aren't working. On my windows machine - no Node.js or VSCode to run it locally

mattpocock (approved):
Working!

davidkpiano (approved):

davidkpiano (approved):

PR #184, opened 2021-08-13
Use Yarn 3 and PnP

Andarist:
PnP scares me a little bit but one thing I really like about this is the ability to switch between branches without having to reinstall dependencies. This would also avoid the install step on CI almost entirely.

Let me know what do u think.


merceyz:
The `.yarn/install-state.gz` file is environment specific and shouldn't be committed
https://yarnpkg.com/getting-started/qa/#which-files-should-be-gitignored

Andarist:
@merceyz thanks for taking a look! I was not aware of that. If we follow on this work I will make sure this gets gitignored properly

PR #188, opened 2021-08-14
Setup Changesets

Andarist:
Screenshot 2021-08-18 at 15 10 22
<details>
<summary>code for the viz above</summary>

```ts
import { createMachine } from 'xstate';
import { raise } from 'xstate/lib/actions';

(createMachine as any)(
  {
    id: 'release workflow',
    initial: 'changesets',
    states: {
      changesets: {
        initial: 'checkout',
        states: {
          checkout: { entry: raise('➡️'), on: { '➡️': 'setup_node' } },
          setup_node: { entry: raise('➡️'), on: { '➡️': 'install_deps' } },
          install_deps: {
            entry: raise('➡️'),
            on: { '➡️': 'create_version_pr' },
          },
          create_version_pr: { type: 'final' },
        },
        onDone: [
          {
            cond: 'hasChangesets',
            target: 'done',
          },
          {
            target: 'create_temp_release_branch',
          },
        ],
      },
      create_temp_release_branch: {
        initial: 'checkout',
        states: {
          checkout: { entry: raise('➡️'), on: { '➡️': 'get_version' } },
          get_version: {
            entry: [raise('➡️')],
            on: { '➡️': 'check_if_release_branch_exists' },
          },
          check_if_release_branch_exists: {
            entry: raise('➡️'),
            on: {
              '➡️': [
                { cond: 'exists', target: '#done' },
                { cond: 'doesnt exist', target: '#deploys' },
              ],
            },
          },
        },
      },
      deploys: {
        type: 'parallel',
        id: 'deploys',
        states: {
          production: {
            initial: 'checkout',
            states: {
              checkout: { entry: raise('➡️'), on: { '➡️': 'setup_git_user' } },
              setup_git_user: {
                entry: raise('➡️'),
                on: { '➡️': 'create_release_commit' },
              },
              create_release_commit: {
                entry: raise('➡️'),
                on: { '➡️': 'force_push_production' },
              },
              force_push_production: {
                entry: raise('➡️'),
                on: { '➡️': 'create_github_release' },
              },
              create_github_release: {
                initial: 'checkout',
                states: {
                  checkout: {
                    entry: raise('➡️'),
                    on: { '➡️': 'publish_tag' },
                  },
                  publish_tag: {
                    entry: raise('➡️'),
                    on: { '➡️': 'publish_release' },
                  },
                  publish_release: {
                    entry: raise('➡️'),
                    on: { '➡️': 'remove_release_branch' },
                  },
                  remove_release_branch: {
                    type: 'final',
                  },
                },
              },
            },
          },
          staging: {
            initial: 'checkout',
            states: {
              checkout: { entry: raise('➡️'), on: { '➡️': 'setup_git_user' } },
              setup_git_user: {
                entry: raise('➡️'),
                on: { '➡️': 'create_release_commit' },
              },
              create_release_commit: {
                entry: raise('➡️'),
                on: { '➡️': 'force_push_staging' },
              },
              force_push_staging: { type: 'final' },
            },
          },
        },
      },
      done: {
        type: 'final',
        id: 'done',
      },
    },
  },
  {
    guards: {
      exists: () => {},
      'doesnt exist': () => {},
    },
  },
);

```
</details>

How this works:
1. whenever we merge to the `dev` branch we either
- create a versioning PR (if there are any changesets on the `dev` branch)
- start the release pipeline (if there are not changesets on the `dev` branch)
2. when versioning PR lands it removes changesets from the `dev` branch which starts the release pipeline like mentioned above
3. the release pipeline prepares 2 jobs - one for releasing staging and one for releasing production
4. in practice, those jobs dont actually make deployments - they just (force) push to appropriare branches so Vercel can pick it up and mak actual deployments
5. those github jobs are not automatic, I've set up reviewers for them in the Environments configuration so they need our approval to execute. That means that we can "approve" them at any given time, using the GitHub interface - the workflow is effectively paused (up to 30 days) until one of us approves it.
6. after approving the production deployment we also create a github release, based on the latest changelog entry

:warn: we should not commit to `staging` and `main` branches - they are just "mirrors" of the scheduled releases, they exist just so they can be used by Vercel. `dev` acts as a single source of truth and it's the only trunk that we use.

There might be times when we need to fix something up quickly - hot fixing on those branches is allowed if such a situation arise. Just keep in mind that you need to separately include any given change in the `dev` branch because the very next release will just overwrite the git history on those deployable branches (cause we are force pushing to them)


.github/workflows/version-or-release.yml
```diff
@@ -0,0 +1,143 @@
...
+          git tag v${{ needs.create-temp-release-branch.outputs.version }}
+          git push origin v${{ needs.create-temp-release-branch.outputs.version }}
+
+      - name: Publish a GitHub release
+        uses: actions/github-script@v4
+        with:
+          script: |
+            const { publishRelease } = require(`./scripts/publish-github-release.js`)
+
+            await publishRelease(github, { repo: context.repo })
```
davidkpiano, line 139: Refactor item: can we put this in a script file?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@xstate/viz",
+  "name": "xstate-viz-app",
```
davidkpiano, line 2: This is internal-only, we won't publish it in the future as this, right?

davidkpiano (approved):

package.json
Andarist: right - i have just changed it to something more distinct, so it shouldnt conflict now with a potential `@xstate/viz` package or smth

.github/workflows/version-or-release.yml
Andarist: Right now the `publish-github-release.js` is more reusable, which won't matter for this one - but in the previous job I've used this pattern to reuse a single script by different workflows, where each workflow was able to parametrize that script how they wanted. If we ever add any such shared script then keeping this one like this makes it more consistent with other, potential, scripts.

In addition to that, providing the `github` (aka `octokit`) allows us to skip installing deps - because we just grab what has been provided to this action and pass and inject it to the script. So it's a small resource optimization

mattpocock (approved):
Looking forward to seeing how it works on dev!

PR #192, opened 2021-08-17
Prioritize direction for initial edges

davidkpiano:
**Before:** Initial state placed at end because ELK is trying to prioritize the edge direction (`RIGHT`) equally for all edges, even the one that cycles back to the initial state, and it is prioritizing the incoming edges instead of the outgoing edges.

CleanShot 2021-08-17 at 17 57 22

**After:** By prioritizing edges whose source is an initial state, they are much more likely to point in the correct direction (`RIGHT`) instead of incoming edges:

CleanShot 2021-08-17 at 17 58 32

Test machine: https://stately.ai/viz?id=746de212-ed0a-44b5-81f7-d4f883bf2f92

### DIFF
--- a/src/Graph.tsx
+++ b/src/Graph.tsx
@@ -136,6 +136,8 @@ function getElkEdge(
     edge,
     sections: [],
     layoutOptions: {
+      // Ensure that all edges originating from initial states point RIGHT
+      // (give them direction priority) so that the initial states can end up on the top left
       'elk.layered.priority.direction': isInitialEdge ? '1' : '0',
     },
   };
--- a/src/Graph.tsx
+++ b/src/Graph.tsx
@@ -110,6 +110,7 @@ function getElkEdge(
   const edgeRect = rectMap.get(edge.id)!;
   const targetPortId = getPortId(edge);
   const isSelfEdge = edge.source === edge.target;
+  const isInitialEdge = edge.source.parent?.initial === edge.source.key;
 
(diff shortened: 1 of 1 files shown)


src/Graph.tsx
```diff
@@ -134,7 +135,9 @@ function getElkEdge(
     ],
     edge,
     sections: [],
-    layoutOptions: {},
+    layoutOptions: {
+      'elk.layered.priority.direction': isInitialEdge ? '1' : '0',
```
Andarist, line 139: could u add a comment here that would go into some details of this solution?

Andarist (approved):

mattpocock (approved):

mattpocock:
Agree with Andarist's comment above

davidkpiano:
**How I figured this out:**

1. Determine that even though direction is `RIGHT`, edges from the initial state were going a different direction, so it's a problem with direction
2. Search Google for "elk direction"
3. Land here: https://www.eclipse.org/elk/reference/options/org-eclipse-elk-direction.html search for other direction options
4. Search for more direction options, land here: https://www.eclipse.org/elk/reference/options/org-eclipse-elk-layered-priority-direction.html which is exactly what I need (direction priority) ✅ 
5. Add direction priority to edges from initial states only, visually verified 🏁

src/Graph.tsx
davidkpiano: Done

PR #194, opened 2021-08-18
Add privacy policy link to footer

davidkpiano:
This PR adds a "more info" button to the canvas view and adds the Privacy Policy link in an easily accessible spot (cc. @laurakalbag)

Also does a bit of cleanup/refactor 🧹 

CleanShot 2021-08-20 at 08 03 44


src/Footer.tsx
```diff
@@ -1,22 +1,20 @@
...
   return (
-    <Box padding="1" paddingInline="2" background="gray.900" gridArea="footer">
-      <Text
-        fontSize="sm"
-        title={JSON.stringify(state.value, null, 2)}
-        color="whiteAlpha.800"
-      >
-        {Array.from(state.tags).join(', ') ?? '&nbsp;'}
-      </Text>
+    <Box
```
Andarist, line 7: IMHO this currently blends with the panels on the right too much. It's also quite big - and quite annoying for me (as a user) because it distracts me. If our app would be mobile-friendly using a footer for this would probably not be an option due to space-limited environment. 

I feel like using a footer for this is a good fit for websites because the footer is not always visible. We have a full-screen application and having this visible at all times is just a waste of space (and a distraction, in my opinion). Especially that this is not something that people need at all times - most won't look into it at all. I've looked through a couple of apps and it seems like putting this in a sidebar is a common practice - it still seems to be very accessible this way, and not too hidden.

davidkpiano: You all really don't like footers 😂  👣  🚫 

Fine, I'll get rid of it.

src/App.tsx
```diff
@@ -1,36 +1,20 @@
...
 import { AuthProvider } from './authContext';
 import { authMachine } from './authMachine';
 import './base.scss';
 import { CanvasProvider } from './CanvasContext';
-import { CanvasPanel } from './CanvasPanel';
-import { toDirectedGraph } from './directedGraph';
-import { EditorPanel } from './EditorPanel';
-import { EventsPanel } from './EventsPanel';
+import { CanvasView } from './CanvasView';
+import { Footer } from './Footer';
```
Andarist, line 9: This PR doesnt compile at the moment because this file doesnt exist anymore.
```suggestion
import { Footer } from './Footer';
```

src/PanelsView.tsx
```diff
@@ -0,0 +1,129 @@
...
+            <EventsPanel />
+          </TabPanel>
+          <TabPanel height="100%" overflowY="auto">
+            <ActorsPanel />
+          </TabPanel>
+          <TabPanel height="100%" overflowY="auto">
+            <SettingsPanel />
+          </TabPanel>
+        </TabPanels>
+        <Box
```
Andarist, line 105: I like the current version better but I still feel like this looks a little bit off:
Screenshot 2021-08-19 at 13 39 57

Maybe it's just me though. It's just that when main buttons when aligned to the bottom of the screen it looked visually better for me.

mattpocock:
I don't love it - especially that the buttons are off the bottom. Could we hide this info behind a ❓  button?

mattpocock:
These buttons should be at the same level

image

davidkpiano:
> I don't love it - especially that the buttons are off the bottom. Could we hide this info behind a ❓ button?

Yep, I'm refactoring to do this.

davidkpiano:
@mattpocock Rereview? (see new image here: https://github.com/statelyai/xstate-viz/pull/194#issue-715273655)

mattpocock:
@davidkpiano Looks good, happy to approve. Could we also add the XState version there as a non-clickable button? And could the button itself have a bg to match the viz bg?

mattpocock (approved):

PR #196, opened 2021-08-19
Fix jumpy editor buttons

Andarist:
It fixes jumpy button that could be seen when the `EditorWithXStateImports` was lazy-loading - it was a minor regression after the Next migration.

https://www.loom.com/share/115bb3391d4b42b0a2544b4fbb49de81


davidkpiano (approved):
Thanks for adding the comment!

Without it, someone might absent-mindedly delete the div in the future, so good call on that 👍

mattpocock (approved):

PR #198, opened 2021-08-19
Added SSR and OG image tags to visualised machines

mattpocock:


src/AppHead.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <meta property="og:image" content={props.ogImageUrl} />
+      )}
+
+      <meta property="twitter:card" content="summary_large_image" />
+      <meta property="twitter:url" content={`https://stately.ai/viz`} />
+      <meta property="twitter:title" content={props.title} />
+      <meta property="twitter:description" content={props.description} />
+      {props.ogImageUrl && (
+        <meta property="twitter:image" content={props.ogImageUrl} />
+      )}
```
Andarist, line 42: Those are somewhat redundant because Twitter already understands og properties. The only thing that we need is `twitter:card`, as that is not part of the OpenGraph standard. 

There is no harm in including both lists but if we can save some bytes - I would be in favor of that :P

src/pages/[sourceFileId].tsx
```diff
@@ -0,0 +1,73 @@
...
+  id: string;
+}
+
+const SourceFileIdPage = (props: SourceFileIdPageProps) => {
+  return (
+    <>
+      <AppHead
+        title={
+          props.data?.name
+            ? `${props.data?.name} | XState Visualiser`
```
Comment (Andarist) [inline src/pages/[sourceFileId].tsx:22]: This is passed to `og:title` "as-is". I'm not entirely sure about this but I think it's usually preferred to keep the `og:title` succinct - without branding etc. It's one of the reasons why `og:title` exists - `<title/>` was already crammed with things like this and the people behind OpenGraph wanted "cleaner" titles

src/AppHead.tsx
mattpocock: Does it? I was under the impression that Twitter needs a different og set, based on my limited research. But if it can be deleted, we should!

src/pages/[sourceFileId].tsx
```diff
@@ -0,0 +1,73 @@
...
+            : `XState Visualiser`
+        }
+        description={
+          props.data?.name ||
+          `Visualizer for XState state machines and statecharts`
```
Comment (Andarist) [inline src/pages/[sourceFileId].tsx:27]: The "Visualizer" in different copies is not consistent - you are using both Visualizer and Visualiser

src/pages/[sourceFileId].tsx
Comment (mattpocock) [reply src/pages/[sourceFileId].tsx:22]: Gotcha, so we should have a different prop for og:title that's cleaner

src/AppHead.tsx
Andarist: My research was also limited 😅  but based on a single source on the internet and vague memory of it being true - it should be supported. We can always add it back if it turns out not be the case

src/sourceMachine.ts
```diff
@@ -123,6 +135,47 @@ export const makeSourceMachine = (auth: SupabaseAuthClient) => {
...
+          states: {
+            checking_if_id_on_query_params: {
+              always: [
+                {
+                  cond: () => {
+                    if (!isOnClientSide()) return false;
+                    const queries = new URLSearchParams(window.location.search);
+
+                    console.log(queries);
+
```
Andarist, line 153:
```suggestion
```

src/pages/[sourceFileId].tsx
Comment (mattpocock) [reply src/pages/[sourceFileId].tsx:27]: Grah

src/AppHead.tsx
mattpocock: Done!

Andarist:
Saving the machine breaks the local development:
https://www.loom.com/share/ddc8416ca0b24dd6bbf100de0701d0f3

mattpocock:
@Andarist I thought I'd fixed this here in `App.tsx`, see the long comment inside `redirectToNewUrlFromLegacyUrl`

mattpocock:
Could you check what your `process.env.NODE_ENV` is set to?

mattpocock:
@Andarist Fixed in latest version

src/sourceMachine.ts
```diff
@@ -123,6 +135,45 @@ export const makeSourceMachine = (auth: SupabaseAuthClient) => {
         },
       },
       states: {
+        checking_if_on_legacy_url: {
```
Andarist, line 138: Wouldnt it be much less involved if we'd just use Next's redirects for that? 🤔 

A config like this could work:
```js
  redirects() {
    return [
      {
        source: '/viz',
        has: [
          {
            type: 'query',
            key: 'id',
            value: '(?<id>.*)',
          },
        ],
        destination: '/viz/:id',
        basePath: false,
        permanent: true,
      },
    ];
  },
```

(I couldn't make it work without `basePath: false`)

This leaves the `id` param in the query though - but that can be just cleaned up client-side, outside of the machine (during the initialization time of the module).

It just seems like such a minor thing and outside of the business logic that externalizing this from the machine seems like a win for me.

mattpocock: Great idea!

src/App.tsx
```diff
@@ -26,7 +33,41 @@ function App() {
...
+     * Apologies for this line of code. The reason this is here
+     * is that XState + React Fast Refresh causes an error:
+     *
+     * Error: Unable to send event to child 'ctx => ctx.sourceRef'
+     * from service 'auth'.
+     *
+     * router.replace causes this in development, but not in prod
+     *
+     * So, we use window.location.href in development (with the /viz
+     * prefix which Next won't automatically) and router.replace in prod
```
Andarist, line 50: isn't some part of the sentence missing here? it looks like this "which Next won't automatically" is cut off

src/App.tsx
```diff
@@ -43,6 +84,8 @@ function App() {
     sourceID,
   });
 
+  if (!isOnClientSide()) return null;
```
Andarist, line 87: So just to get this right - when SSRing we actually don't render anything (btw. this results in React warning that gets logged in the console when rehydrating). Which parts of this component are relevant for SSR then?

src/AppHead.tsx
```diff
@@ -0,0 +1,48 @@
+import Head from 'next/head';
+
+interface AppHeadProps {
+  title: string;
+  /**
+   * Og titles are expected to be more concise,
+   * whereas titles in the <title> attribute
+   * should be SEO-friendly
+   *
+   * ogTitle: My Great Machine
```
Andarist, line 10: With this
```suggestion
   * @example
   * ogTitle: My Great Machine
```
it can render slightly better in the IDE.

**Before**
Screenshot 2021-08-24 at 14 02 49
**After**
Screenshot 2021-08-24 at 14 02 34

cypress/support/commands.ts
```diff
@@ -73,6 +74,33 @@ const visitInspector = () => {
...
+ * its SSR return
+ */
+const visitVizWithNextPageProps = (
+  data: Partial<GetSourceFileSsrQuery['getSourceFile']> & { id: string },
+) => {
+  cy.visit(`/viz/${data.id}?noSSR`, {
+    onBeforeLoad: (win) => {
+      let nextData: any;
+
+      Object.defineProperty(win, '__NEXT_DATA__', {
```
Andarist, line 88: This is a neat solution to the problem. 👍 "Abusing" getters & setters (and other fun features of JS/browsers) like this is my jam 🤣 

I was wondering about potential downsides though and something that comes to mind is that this still results in an incorrect HTML - right now we don't really utilize it that much so it works OK but, in the future, this might be problematic because the true response is 404 and it won't hold the "correct" HTML. We are also still rendering the component in question - just with empty/stub props. Which can easily break if we start relying on some new data etc.

I was wondering a bit about alternatives that would not suffer from those problems and I think I've come up with a neat solution - that wouldn't be hard to implement. It took me some time to come up with this but it seems like a **super** easy solution could be leveraged:
```js
cy.visit(`/viz/${data.id}?ssr=${JSON.stringify(data)}`)
```
And just use the content of the `ssr` param on the other side to render what we need. A nice trait about this is that it should be completely type-safe, without the need for any casting etc (apart from the fact that we are going to `JSON.parse` on the other side)

cypress/support/commands.ts
```diff
@@ -73,6 +74,33 @@ const visitInspector = () => {
...
+            id: data.id,
+          };
+          nextData = o;
+        },
+        get() {
+          return nextData;
+        },
+      });
+    },
+  });
```
Andarist, line 101: This is a neat solution to the problem. 👍 "Abusing" getters & setters (and other fun features of JS/browsers) like this is my jam 🤣 

I was wondering about potential downsides though and something that comes to mind is that this still results in an incorrect HTML - right now we don't really utilize it that much so it works OK but, in the future, this might be problematic because the true response is 404 and it won't hold the "correct" HTML. We are also still rendering the component in question - just with empty/stub props. Which can easily break if we start relying on some new data etc.

I was wondering a bit about alternatives that would not suffer from those problems and I think I've come up with a neat solution - that wouldn't be hard to implement. It took me some time to come up with this but it seems like a **super** easy solution could be leveraged:
```js
cy.visit(`/viz/${data.id}?ssr=${JSON.stringify(data)}`);
```
And just use the content of the `ssr` param on the other side to render what we need. A nice trait about this is that it should be completely type-safe, without the need for any casting etc (apart from the fact that we are going to `JSON.parse` on the other side)

src/ShareButton.tsx
```diff
@@ -12,58 +12,87 @@ import {
...
+                  twitterText,
+                )}`,
+              );
+            }}
+          >
             <HStack spacing="3">
               <Twitter fill="white" />
               <Text>Twitter</Text>
             </HStack>
           </MenuItem>
```
Andarist, line 78: It would be good to use a Link here - it's a nice feature of a browser that the target URL gets displayed in a corner. I've realized that I'm actually often looking at this. It's also definitely more semantic and might also come with positive a11y improvements (although I'm not super sure about it as the default `role` of `<a/>` gets overridden here anyway)

```suggestion
          <MenuItem
            as={Link}
            href={`https://twitter.com/intent/tweet?text=${encodeURIComponent(
              twitterText,
            )}`}
            target="_blank"
            rel="noopener noreferrer"
          >
            <HStack spacing="3">
              <Twitter fill="white" />
              <Text>Twitter</Text>
            </HStack>
          </MenuItem>
```

Note: this might require some minor styling adjustments - we might not want to have underline here and similar things.

.changeset/tiny-chicken-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"xstate-viz-app": patch
+---
+
+Added SSR and OG image tags to visualised machines, and put visualised machines on a new, prettier URL - `/viz/:id`
```
Andarist, line 5: All 3 things are completely separate and unrelated from the user's PoV. Don't be afraid of creating more than a single changeset per PR :)

I'm not sure if SSR is that interesting to our users - but the other ones definitely are! A sample OG image could also be attached to a changeset mentioning this improvement.

src/App.tsx
mattpocock: Correct. But not sure I understand the question?

src/App.tsx
mattpocock:
```suggestion
     * So, we use window.location.href in development (with the /viz
     * prefix which Next won't automatically add) and router.replace in prod
```

src/App.tsx
Andarist: Is it important for SSR to execute anything in this component? I mean - could the `isOnClientSide` check be moved to the top of this component? 

And also - what we can do about React warning?

src/sourceMachine.ts
mattpocock: @Andarist This results in a URL which looks like this:

`http://localhost:3000/viz/a3613bbd-aeff-4618-a7e7-08d11395ec79?id=a3613bbd-aeff-4618-a7e7-08d11395ec79`

Which I suppose is fine

src/App.tsx
mattpocock: We could move it to above this component in the tree? I.e. `<NoSSR><App /></NoSSR>`. I don't like this because I like the idea you could use `<App />` without caring if it needs to be SSR'd or not.

The reason it's at the bottom is so we don't conditionally render hooks.

We could make a HOC?

`withNoSSR(App)`

But I prefer the above implementation.

cypress/support/commands.ts
mattpocock: @Andarist I'm guessing we'd need to encodeURIComponent on that data too? Will try it now.

mattpocock: Works great!

src/sourceMachine.ts
Andarist: @mattpocock 

> This leaves the id param in the query though - but that can be just cleaned up client-side, outside of the machine (during the initialization time of the module).

src/App.tsx
Andarist: > The reason it's at the bottom is so we don't conditionally render hooks.

Forgot about this pesky thing. It's fine how it is then.

src/sourceMachine.ts
Andarist: I don't care about this being removed or not though, like you say - it rly doesn't matter

src/pages/[sourceFileId].tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const sourceFileId = req.query.sourceFileId as string;
+    const result = await gQuery(GetSourceFileSsrDocument, {
+      id: sourceFileId,
+    });
+
+    if (!result.data?.getSourceFile) {
+      return {
+        notFound: true,
+        props: {} as any,
```
Comment (Andarist) [inline src/pages/[sourceFileId].tsx:55]: It typechecks just fine without it:
```suggestion
```

The `notFound` is already included in the union here:
https://github.com/vercel/next.js/blob/53f0973abaa5db4961dc03b226698d3c4ddbcf2b/packages/next/types/index.d.ts#L162-L165

Andarist (approved):

PR #186, opened 2021-08-13
Move event to useEffect and add e.preventDefault()

santicros:
This PR fixes #185, where zooming on the canvas with the trackpad wheel + CMD or CTRL zoomed the whole page.

Adding e.preventDefault() on the current onWheel did not work, so I had to move the eventListener to a new useEffect.

Before this PR:
https://user-images.githubusercontent.com/24988424/129388288-72ab1939-e0f9-4593-a1d4-734b380906b9.mov

After this PR:
https://user-images.githubusercontent.com/24988424/129389217-9c6d98f9-7170-46c1-83a5-e708601f411b.mov

PR #200, opened 2021-08-20
Ensure that all keyboard shortcuts are using platform-specific meta key

Andarist:
This PR targets my other PR that has implemented this change for one of our wheel handlers. That PR didn't land yet on `dev` so I've prepared this one as a separate one - just to make review process easier. Once approved I will merge this into that other branch and land them both on dev, at once.

closes STA-153

### DIFF
--- a/src/CanvasContainer.tsx
+++ b/src/CanvasContainer.tsx
@@ -4,7 +4,7 @@ import { useCanvas } from './CanvasContext';
 import { useMachine } from '@xstate/react';
 import { createModel } from 'xstate/lib/model';
 import { Point } from './pathUtils';
-import { isMac } from './utils';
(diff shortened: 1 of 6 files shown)


mattpocock (changes requested):
Wheel still doesn't work on Windows. Ctrl-K on windows brings up command palette, but also focuses search bar. Ctrl-S works as expected

Andarist:
@mattpocock fixed the issue with ctrl+K, nice catch! as to the wheel - this is not part of this PR, the fix for it is here: https://github.com/statelyai/xstate-viz/pull/186

mattpocock (approved):
Approved after further testing

davidkpiano (approved):

PR #203, opened 2021-08-20
Fixed an issue with editor buttons being left outside of the screen after viewport being downsized

Andarist:
This has been broken... once again 🤪 

This is an attempt to fix the issue once and for all - inspired by https://twitter.com/souporserious/status/1427404212531458054

### DIFF
--- a/cypress/integration/visual.spec.ts
+++ b/cypress/integration/visual.spec.ts
@@ -1,18 +1,46 @@
-const getCenterY = (el: HTMLElement) => {
-  const { y, height } = el.getBoundingClientRect();
+const getCenterY = ($el: JQuery<HTMLElement>) => {
+  const { y, height } = $el.get(0).getBoundingClientRect();
   return y + height / 2;
 };
 
+const isInViewport = ($el: JQuery<HTMLElement>) => {
+  const { fromElViewport } = Cypress.dom.getElementPositioning($el);
+  // Cypress types are broken for this at the moment
+  const win = Cypress.dom.getWindowByElement($el.get(0)) as any as Window;
+
+  return (
+    fromElViewport.top >= 0 &&
+    fromElViewport.right <= win.innerWidth &&
+    fromElViewport.bottom <= win.innerHeight &&
+    fromElViewport.left >= 0
+  );
+};
+
 describe('Visual', () => {
+  it('should have editor buttons in the viewport after the viewport gets downsized', () => {
+    cy.visit('/viz');
+
+    cy.getMonacoEditor();
+
+    cy.viewport(
+      Cypress.config().viewportWidth,
+      Cypress.config().viewportHeight - 300,
+    );
+
+    cy.findByRole('button', { name: 'Visualize' }).then(($button) => {
+      expect(isInViewport($button)).to.be.true;
+    });
+  });
+
   it('should have editor buttons aligned vertically with canvas buttons', () => {
     cy.visit('/viz');
 
     cy.findByRole('button', { name: 'Visualize' }).then(($visualize) => {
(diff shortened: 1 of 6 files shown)


src/paletteMachine.ts
```diff
@@ -24,8 +23,7 @@ export const paletteMachine = paletteModel.createMachine({
             return Object.values(keybindings).some(Boolean);
           }
           function eventRoseFromEditor(e: KeyboardEvent) {
-            const editorElement = document.querySelector(EDITOR_CLASSNAME);
-            return editorElement && editorElement.contains(e.target as Node);
+            return !!(e.target as Element).closest('.monaco-editor');
```
Andarist, line 26: This fixes a small additional issue - now each monaco will be detected by this check, including the one for JSON payloads. So pressing Shift+? won't open the palette when you write in that monaco instance

src/theme.ts
```diff
@@ -13,6 +13,8 @@ export const theme = extendTheme({
     global: {
       '*, *:before, *:after': {
         position: 'relative',
+        minHeight: 0,
+        minWidth: 0,
```
Andarist, line 17: Might need to reconsider this approach - somebody has mentioned a drawback here: https://twitter.com/chungwu/status/1427413289621876743

src/paletteMachine.ts
Andarist: I've moved this fix (and improved it) to https://github.com/statelyai/xstate-viz/pull/208 . Gonna remove the related changes from this PR in a sec

src/EditorWithXStateImports.tsx
```diff
@@ -105,96 +104,85 @@ export const EditorWithXStateImports = (
...
-                        text: prettify(editor.getValue()),
-                        range: model.getFullModelRange(),
-                      },
-                    ];
-                  } catch (err) {
-                    console.error(err);
-                  } finally {
-                    props.onFormat?.();
-                  }
+    <Editor
```
davidkpiano, line 107: The only change here is the removal of `<ClassNames>`, right?

davidkpiano (approved):

src/EditorWithXStateImports.tsx
Andarist: Ye, gonna try to remember to mention this kind of stuff in the PR comments next time 👍

mattpocock (approved):

cypress/integration/visual.spec.ts
```diff
@@ -1,18 +1,46 @@
...
+    cy.getMonacoEditor();
+
+    cy.viewport(
+      Cypress.config().viewportWidth,
+      Cypress.config().viewportHeight - 300,
+    );
+
+    cy.findByRole('button', { name: 'Visualize' }).then(($button) => {
+      expect(isInViewport($button)).to.be.true;
+    });
```
mattpocock, line 32: Question - can't we just `cy.click` this? Cypress should fail if the button isn't in the viewport

davidkpiano (approved):

cypress/integration/visual.spec.ts
mattpocock: No need to hold up the PR for this though, just curious

Andarist: Clicking doesn't work correctly - it automatically scrolls to the element (although it can be disabled with a flag, AFAIR) and focused the element (which makes the browser forcefully scroll to the element). The first isn't as much of a problem for us but still something that could blur the test scenario.

If I try to use `cy.click()` I end up with a state like this:
Screenshot 2021-08-30 at 18 41 43

Note that this is without the fix being applied - the browser (most likely) has shifted the whole thing so the button became part of the viewport.

PR #206, opened 2021-08-20
Refactor spacing around bottom buttons in order to avoid using a fixed height on the button container

Andarist:
Was trying to achieve the same visual effect without the fixed heights, as a CSS exercise for myself. Got the desired effect - so I thought it might be worth creating a PR with this. It feels like this is a spacing problem and as such a spacing-oriented solution seems to be more in the spirit of design systems 🤔

### DIFF
--- a/src/EditorPanel.tsx
+++ b/src/EditorPanel.tsx
@@ -426,7 +426,7 @@ export const EditorPanel: React.FC<{
           }}
         />
       )}
-      <Box height="100%" display="grid" gridTemplateRows="1fr 4rem">
+      <Box height="100%" display="grid" gridTemplateRows="1fr auto">
         {simulationMode === 'visualizing' && (
           <>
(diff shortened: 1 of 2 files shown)


mattpocock:
🤷  Sometimes fixed heights are more predictable to work with? I don't have a strong feeling about this PR, it looks the same but I have found fixed heights to be a perfectly acceptable approach. Shrug & Approve

mattpocock (approved):

src/EditorPanel.tsx
```diff
@@ -426,7 +426,7 @@ export const EditorPanel: React.FC<{
           }}
         />
       )}
-      <Box height="100%" display="grid" gridTemplateRows="1fr 4rem">
+      <Box height="100%" display="grid" gridTemplateRows="1fr auto">
```
davidkpiano, line 429: This might not be good if `auto` becomes a different size than the other footer (buttons will be misaligned)

PR #208, opened 2021-08-21
Fixed an issue with global keybindings being also triggered when providing input to input-like elements

Andarist:
This moves a fix from here: https://github.com/statelyai/xstate-viz/pull/203#discussion_r693093759 and also makes it better because this was only handling the Monaco editors. This PR fixes the problem for all kind of input-like elements. In our application this includes:
- event payload's JSON Monaco
- event list filter input
- name chooser input in a modal

### DIFF
--- a/src/paletteMachine.ts
+++ b/src/paletteMachine.ts
@@ -1,6 +1,5 @@
 import { createModel } from 'xstate/lib/model';
-import { EDITOR_CLASSNAME } from './constants';
-import { isWithPlatformMetaKey } from './utils';
+import { isWithPlatformMetaKey, isInputLikeElement } from './utils';
 
 const paletteModel = createModel(undefined, {
   events: {
@@ -23,14 +22,10 @@ export const paletteMachine = paletteModel.createMachine({
             };
             return Object.values(keybindings).some(Boolean);
           }
-          function eventRoseFromEditor(e: KeyboardEvent) {
-            const editorElement = document.querySelector(EDITOR_CLASSNAME);
-            return editorElement && editorElement.contains(e.target as Node);
-          }
           const eventHandler = (e: KeyboardEvent) => {
             if (
-              captureCommandPaletteKeybindings(e) &&
-              !eventRoseFromEditor(e)
+              !isInputLikeElement(e.target as HTMLElement) &&
+              captureCommandPaletteKeybindings(e)
             ) {
               e.preventDefault();
               sendBack('SHOW_PALETTE');
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -202,3 +202,6 @@ export function isDelayedTransitionAction(
       return false;
   }
 }
+
+export const isInputLikeElement = (el: HTMLElement) =>
+  el.tagName === 'INPUT' || el.tagName === 'TEXTAREA' || el.isContentEditable;
--- a/.changeset/strange-ligers-walk.md
+++ b/.changeset/strange-ligers-walk.md
@@ -1,0 +1,5 @@
+---
+'xstate-viz-app': patch
+---
+
+Fixed an issue with global keybindings being also triggered when providing input to input-like elements.
--- a/src/CanvasContainer.tsx
+++ b/src/CanvasContainer.tsx
@@ -4,7 +4,7 @@ import { useCanvas } from './CanvasContext';
 import { useMachine } from '@xstate/react';
 import { createModel } from 'xstate/lib/model';
 import { Point } from './pathUtils';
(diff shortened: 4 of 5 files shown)


src/utils.ts
```diff
@@ -202,3 +202,6 @@ export function isDelayedTransitionAction(
       return false;
   }
 }
+
+export const isInputLikeElement = (el: HTMLElement) =>
+  el.tagName === 'INPUT' || el.tagName === 'TEXTAREA' || el.isContentEditable;
```
Andarist, line 207: TIL I've learned that there is `.isContentEditable` property that you can check and it works for all elements within `[contendeditable]`

src/paletteMachine.ts
```diff
@@ -23,14 +22,10 @@ export const paletteMachine = paletteModel.createMachine({
...
           }
-          function eventRoseFromEditor(e: KeyboardEvent) {
-            const editorElement = document.querySelector(EDITOR_CLASSNAME);
-            return editorElement && editorElement.contains(e.target as Node);
-          }
           const eventHandler = (e: KeyboardEvent) => {
             if (
-              captureCommandPaletteKeybindings(e) &&
-              !eventRoseFromEditor(e)
+              !isInputLikeElement(e.target as HTMLElement) &&
```
Andarist, line 27: it means that the palette also won't be showed when we have focus on checkboxes, radios and similar - as the check inside this helper only checks the `tagName`. This seems fine though? Maintaining a list of allowlist/denylist of input types seemed a little bit painful for me so I've ignored this for the time being. Thoughts?

davidkpiano: I think it should show. What if we changed our approach to stopping propagation inside inputs, rather than checking if the event came from an input?

Andarist: I've considered stopping the propagation approach in the past but the problem with that is that then the "logic" for that is scattered all over the place and it's easy to regress on the correct behaviour. You need to keep an eye on this and be aware that this is even a thing - whereas this logic is centralized and when you add new components to your app you simply don't have to care about this issue.

I've done a summary for all `input` types to check out how they behave and what kind of a check could be implemented here:
| `input` type   | keyboard input                                                                        |
| -------------- | ------------------------------------------------------------------------------------- |
| button         | definitely doesn't accept text                                                        |
| checkbox       | definitely doesn't accept text                                                        |
| color          | it renders as a button - keydowns in the popup are ignored                            |
| date           | it only accepts numerical text - keydowns in the popup are ignored                    |
| datetime-local | it only accepts numerical text - keydowns in the popup are ignored                    |
| email          | definitely accepts text                                                               |
| file           | it renders as a button                                                                |
| hidden         | it doesn't render at all                                                              |
| image          | it acts as a graphical submit button                                                  |
| month          | it only accepts alphanumerical text - keydowns in the popup are ignored               |
| number         | it only accepts numerical text                                                        |
| password       | definitely accepts text                                                               |
| radio          | definitely doesn't accept text                                                        |
| range          | it only accepts arrows                                                                |
| reset          | it renders as a button                                                                |
| search         | definitely accepts text                                                               |
| submit         | it renders as a button                                                                |
| tel            | definitely accepts text                                                               |
| text           | definitely accepts text                                                               |
| time           | it only accepts numerical text and spacebar/enter - keydowns in the popup are ignored |
| url            | definitely accepts text                                                               |
| week           | it only accepts numerical text - keydowns in the popup are ignored                    |

The intention was not to list all meaningful keyboard interactions for those but to give a vague notion of what kind of an input is accepted by them. I was testing this here: https://codesandbox.io/s/kind-browser-9hrwz?file=/src/App.js

I think we only have two available strategies here to allow those interactions on some inputs - it's either to allow them for specific input types or to disallow them for the complementary values. The first one seems to be conceptually slightly better - if we miss something or a new input type gets added to the platform the worst thing that happens is that the shortcut won't work in those cases. Using the opposite strategy bears the risk of interfering with the user input - which is a bad UX.

It's really tempting to go with the second approach though as the list of input types that would have to be used for it is much shorter.

It's also worth noting that this simple approach only works for certain global shortcuts - we could imagine a situation in other apps when a particular shortcut would clash with, for example, numerical inputs. In such a situation I imagine one would have to implement different strategies for different in-app shortcuts that would account for different input types. This doesn't apply to us though - at least not right now.

I've implemented adjustments here: https://github.com/statelyai/xstate-viz/pull/208/commits/2525df88c888c68d61eaec6061610a29b4236ccc

mattpocock (approved):
Tested, working well!

davidkpiano (approved):
LGTM; is there an easy way to write tests for this? (Okay if not)

PR #209, opened 2021-08-22
Embedded mode

farskid:
Adds embedded mode.

### Defaults:

param|default|description|variants
---|---|---|---
mode|viz|the embedded mode|`viz`,`panels`,`full`
panel|code|the active panel in the right-side panel|`code`,`state`,`events`,`actors`
showOriginalLink|1|whether to show the button that opens the original viz|`0`,`1`
readOnly|1|whether code editor is read-only or not|`0`,`1`
pan|0|whether panning in canvas should be allowed|`0`,`1`
zoom|0|whether zooming in/out on canvas should be allowed|`0`,`1`
controls|0|whether control buttons should be shown|`0`,`1`

#### mode:`viz`
image

#### mode:`panels` and panel:`actors`
image

### mode:`panels`, panel:`actors` and showOriginalLink:`0`
image

### mode:`full` and panel:`actors`
image

### mode:`full`, panel:`actors` and readOnly:`1`
image


mattpocock:
@farskid Could you fix the merge conflicts? I'm also seeing a 500 from Vercel on the preview deploy

farskid:
> @farskid Could you fix the merge conflicts? I'm also seeing a 500 from Vercel on the preview deploy

Will do it as soon as I get the last commit pushed.

src/canvasMachine.tsx
```diff
@@ -105,8 +105,8 @@ const getNewZoomAndPan = (
   };
 };
 
-export const canvasMachine = createMachine<typeof canvasModel>({
-  context: canvasModel.initialContext,
+export const canvasMachine = canvasModel.createMachine({
+  // context: canvasModel.initialContext,
```
davidkpiano, line 109:
```suggestion
```

davidkpiano:
Bikeshedding: should we call `full` mode `split` instead? (poor naming on my part)

src/Visibility.tsx
```diff
@@ -0,0 +1,14 @@
+import { Box } from '@chakra-ui/react';
+import React from 'react';
+
+export const Visibility: React.FC<
+  {
+    isHidden: any;
```
Andarist, line 6:
```suggestion
    isHidden: boolean;
```

src/PanelsView.tsx
```diff
@@ -21,28 +23,50 @@ import { useSourceActor } from './sourceMachine';
...
+      gridArea="panels"
+      minHeight={0}
+      hidden={embed.isEmbedded && embed.mode === 'viz'}
+    >
       <Tabs
         bg="gray.800"
         display="grid"
         gridTemplateRows="3rem 1fr"
         height="100%"
+        defaultIndex={embed.panelIndex}
```
Andarist, line 44: `panelIndex` is only used here - IMHO it would make sense to keep it local and just compute/derive it here

src/pages/index.tsx
```diff
@@ -1,12 +1,29 @@
...
-  return null;
+const HomePage = ({ query }: { query: any }) => {
+  const embed = parseEmbedQuery(query);
+  return (
+    <App
+      embed={{
+        ...embed,
+        isEmbedded: false,
+        embedUrl: null,
+      }}
```
Andarist, line 20: This is a little bit strange to me - I don't see how `embed` prop is relevant for a non-embed route. This makes me think that this is really not an embed configuration but rather a set of feature flags, available for any route. Which is fine - but then I would definitely not call this `embed` because it makes the thing confusing

src/pages/index.tsx
```diff
@@ -1,12 +1,29 @@
...
+    />
+  );
+};
+
+export const getServerSideProps: GetServerSideProps = async (ctx) => {
+  return { props: { query: ctx.query } };
 };
```
Andarist, line 27: note: at the moment we don't do "true" SSR so this doesn't quite have to include `getServerSideProps` (the query will be available client-side anyway). By removing that we'd be able to optimize this because this index route could be statically generated - whereas right now it has to be server-side rendered.

OTOH if we ever start to truly SSR this then this will become required.

src/pages/embed.tsx
```diff
@@ -0,0 +1,21 @@
...
+      embed={{
+        ...embed,
+        isEmbedded: true,
+        embedUrl: withoutEmbedQueryParams(query),
+      }}
+    />
+  );
+}
+
+Embed.getInitialProps = async (ctx: NextPageContext) => {
```
Andarist, line 18: you might have wanted [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) - `getInitialProps` is sort of deprecated/fading away

src/ResizableBox.tsx
```diff
@@ -106,16 +107,26 @@ const ResizeHandle: React.FC<{
 
 export const ResizableBox: React.FC<Omit<BoxProps, 'width'>> = ({
   children,
+  hidden,
   ...props
 }) => {
+  const { isEmbedded } = useEmbed();
```
Andarist, line 113: `ResizableBox` sounds like a quite generic component, decoupled from our business logic - this makes me think that reading this state here should not be its responsibility. It could accept a `disabled` prop or something instead of that

src/PanelsView.tsx
```diff
@@ -21,28 +23,50 @@ import { useSourceActor } from './sourceMachine';
...
           <Tab>State</Tab>
           <Tab>Events</Tab>
           <Tab>Actors</Tab>
-          <Tab marginLeft="auto" marginRight="2">
+          <Tab marginLeft="auto" marginRight="2" hidden={embed.isEmbedded}>
             <SettingsIcon />
           </Tab>
-          <Login />
+          <Login hidden={embed.isEmbedded} />
+          {embed.isEmbedded && embed.showOriginalLink && embed.embedUrl && (
```
Andarist, line 55: `showOriginalLink` and `embedUrl` look a little bit confusing together here, should the latter be `originalUrl` (and the former could become `showOriginalUrl`?

src/PanelsView.tsx
```diff
@@ -98,7 +122,7 @@ export const PanelsView = () => {
           <TabPanel height="100%" overflowY="auto">
             <ActorsPanel />
           </TabPanel>
-          <TabPanel height="100%" overflowY="auto">
+          <TabPanel height="100%" overflowY="auto" hidden={embed.isEmbedded}>
```
Andarist, line 125: this could not only be hidden - this could not even be rendered (based on this condition)

src/PanelsView.tsx
```diff
@@ -21,28 +23,50 @@ import { useSourceActor } from './sourceMachine';
...
         height="100%"
+        defaultIndex={embed.panelIndex}
       >
         <TabList>
           <Tab>Code</Tab>
           <Tab>State</Tab>
           <Tab>Events</Tab>
           <Tab>Actors</Tab>
-          <Tab marginLeft="auto" marginRight="2">
+          <Tab marginLeft="auto" marginRight="2" hidden={embed.isEmbedded}>
```
Andarist, line 51: https://github.com/statelyai/xstate-viz/pull/209/files#r700863257

src/types.ts
```diff
@@ -46,3 +46,32 @@ export type SimulationMode = 'inspecting' | 'visualizing';
...
+export enum EmbedPanel {
+  Code = 'code',
+  State = 'state',
+  Events = 'events',
+  Actors = 'actors',
+  Settings = 'settings',
+}
+export interface ParsedEmbed {
+  mode: EmbedMode;
+  panel: EmbedPanel;
```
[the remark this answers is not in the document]
farskid: Yes and I wanted the naming to mirror the name of the query parameter.

src/pages/index.tsx
farskid: it's fixed in the last changes.

src/pages/index.tsx
farskid: this needs to be read on the server-side because otherwise, the embedded mode will first load normally and then `useRouter` will pick up available query params and UI will have a flaky jump between the 2 modes.

src/pages/embed.tsx
farskid: fixed in the latest changes.

src/ResizableBox.tsx
farskid: I agree. fixing

cypress/integration/embedded-mode.spec.ts
```diff
@@ -1,64 +1,51 @@
-const SOURCE_ID = 'c75ccdc4-418d-4ade-9059-90f0fc4ddbd1';
+const SOURCE_ID = 'fc126bd2-b903-4777-b7f4-c1c33410cc21';
```
mattpocock, line 1: Is this checking a source in the viz itself? This edges towards this being an e2e test, I'd prefer it if we mocked this.

farskid: what's the best way of mocking it?

src/PanelsView.tsx
farskid: What do you mean?

davidkpiano: I think he's saying that we don't have to render this tab at all if embedded

farskid:
> image
> 
> When accessing `/viz`

This is fixed now.

src/PanelsView.tsx
farskid: right. I just kept everything as hidden because, in some places, we send events to some actors that affect the entire logic. Though keeping everything but hiding, for now, might be a safe trade-off until we take those bits of logic out of React.

src/PanelsView.tsx
farskid: https://github.com/statelyai/xstate-viz/pull/209#discussion_r701576538

farskid:
an example of embedded mode meets inspector mode:
https://codesandbox.io/s/vigorous-rosalind-4ysr9?file=/src/index.tsx

src/PanelsView.tsx
farskid: ok fair enough. I changed all hidden elements to be conditionally excluded from the DOM.

davidkpiano:
Getting this error (might be something wrong on my end?)

CleanShot 2021-09-03 at 14 32 33

farskid:
> Getting this error (might be something wrong on my end?)
> 
> CleanShot 2021-09-03 at 14 32 33

what page do you get this on?

mattpocock:
@farskid Is there a way to remove the `?`? Should this be a part of the 'controls' flag?

farskid:
> @farskid Is there a way to remove the Question Mark Icon? Should this be a part of the 'controls' flag?

@mattpocock I suppose it could be. what's the use case?

mattpocock:
> @mattpocock I suppose it could be. what's the use case?

For OG image mode. Don't want that question mark there for that.

farskid:
> > @mattpocock I suppose it could be. what's the use case?
> 
> @farskid For OG image mode. Don't want that question mark there for that.

@mattpocock How about the RESET button? wanna hide that too?

mattpocock:
Yes! Sounds good.

cypress/integration/saving.spec.ts
```diff
@@ -1,26 +1,17 @@
...
-      },
-    };
-
-    cy.intercept('source-file-id.json?sourceFileId=source-file-id', {
-      pageProps: {
-        id: sourceFileToBeCreated.id,
-        data: sourceFileToBeCreated,
-      },
-    });
     cy.interceptGraphQL({
```
mattpocock, line 4: What are the reasons for the changes in this file?

src/ExternalIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export const ExternalIcon: React.FC = () => (
```
mattpocock, line 3: Nit: should probably add this to the Icons list

src/pages/_app.tsx
```diff
@@ -61,12 +67,22 @@ const MyApp = ({ pageProps, Component }: AppProps) => {
...
   return (
     <AuthProvider value={authService}>
-      <Component {...pageProps} />
+      <EmbedProvider
+        value={{
+          ...embed,
+          isEmbedded: pageProps.isEmbedded,
+          originalUrl: withoutEmbedQueryParams(pageProps.query),
+        }}
+      >
```
mattpocock, line 81: What's the reason for providing this at the root level? The auth service makes sense to be at the root, but the embed page might be better to be placed inside `App.tsx` - the idea being that we may end up having routes in the viz that don't care about the monaco editor or whether things are embedded or not.

src/pages/embed/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (ctx.query.inspect === undefined) {
+    return {
+      notFound: true,
+      props: {},
+    };
+  }
+  return {
+    props: { query: ctx.query, isEmbedded: true },
+  };
+};
```
mattpocock, line 41: I think we should remove this and handle this client-side if we can. Every `getServerSideProps` is a lambda call, and we should cut down on that as much as possible. All the logic in here can be handled with `useRouter`

src/pages/index.tsx
```diff
@@ -1,6 +1,14 @@
...
 import App from '../App';
 import { AppHead } from '../AppHead';
 
+/**
+ * This is a shim for now. Our app currently doesn't use
+ * next routing - we're just loading the entire app in _app.tsx
+ *
+ * This will be fixed in a later PR, adding routes with server
+ * rendering to grab OG images.
+ */
```
mattpocock, line 11: Is this still true? Don't think it is.

src/sourceMachine.ts
```diff
@@ -141,11 +142,27 @@ class NotFoundError extends Error {
   }
 }
 
+function getInvocations(isEmbedded: boolean) {
```
mattpocock, line 145:
```suggestion
// TODO - find a better way to handle this than dynamically changing the invoked services
function getInvocations(isEmbedded: boolean) {
```

src/sourceMachine.ts
```diff
@@ -286,8 +303,14 @@ export const makeSourceMachine = (params: {
                     assign({
                       sourceRawContent: (ctx, e) => e.code,
                     }),
-                    forwardTo('codeCacheMachine'),
-                    forwardTo('confirmBeforeLeavingMachine'),
+                    pure(() => {
+                      if (!params.isEmbedded) {
```
mattpocock, line 307: Nit: Maybe we can handle this in a cond? Or a choose? Something more visualisable.

src/pages/_app.tsx
farskid: The reason I put it in _app was that your pages either are embedded or not but both need to pass on some metadata even when it's not the embedded page. So each page had to deal with passing that data manually where putting it in the _app will do it once for all pages.

src/pages/embed/index.tsx
farskid: We can't do it on the client-side. `useRouter` first only receives the query params when it's late (after the 1st paint) so you'll see a flash of jumping layout when done in the client side.

src/pages/index.tsx
farskid: I think not.

src/ExternalIcon.tsx
farskid: How would you handle different viewbox?

cypress/integration/saving.spec.ts
farskid: Not sure why these are removed. Could've been a mistake when we were trying to fix the cypress tests.

src/pages/_app.tsx
mattpocock: Why not in `App.tsx`, though? That is acting as a wrapper for everything we'd need embedded mode for - the Monaco editor, canvas etc.

src/ExternalIcon.tsx
mattpocock: There's a `makeIcon` function in there that allows you to pass a different viewBox

src/pages/embed/index.tsx
mattpocock: > We can't do it on the client-side. `useRouter` first only receives the query params when it's late (after the 1st paint) so you'll see a flash of jumping layout when done in the client side.

Can we make a wrapper for it to check the query params first? We can probably manage the flash of jumping layout by returning null for the first render inside `App.tsx`

farskid: We can do that I suppose. we probably still need to pass `isEmbedded` as a prop to `App` to avoid parsing the url.

mattpocock (approved):
I'm happy to send this to staging for further testing

PR #211, opened 2021-08-24
Recoloured viz based on user feedback

mattpocock:
image

This deprioritises actions you can't trigger, adds a faded out blue border to states to indicate what they are, and deepens the primary blue to increase accessibility.

### DIFF
--- a/src/StateNodeViz.scss
+++ b/src/StateNodeViz.scss
@@ -64,7 +64,7 @@
 
 [data-viz='stateNode-content'] {
   background: var(--viz-node-color-bg);
-  padding: 0.6rem 1.4rem;
+  padding: 0.5rem 1.5rem;
 
   &:empty {
     display: none;
--- a/src/StateNodeViz.scss
+++ b/src/StateNodeViz.scss
@@ -64,7 +64,7 @@
 
 [data-viz='stateNode-content'] {
   background: var(--viz-node-color-bg);
-  padding: 1rem 2rem;
+  padding: 0.6rem 1.4rem;
 
   &:empty {
     display: none;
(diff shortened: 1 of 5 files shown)


src/TransitionViz.scss
```diff
@@ -1,12 +1,15 @@
 [data-viz='transition'] {
-  --viz-transition-color: gray;
+  --viz-transition-color: #333;
   display: block;
-  border-radius: 1rem;
+  border-radius: 0.4rem;
```
davidkpiano, line 4:
```suggestion
  border-radius: 1rem;
```

I feel that we should have transitions visually distinct from the states in overall shape, with more rounded corners 🤔

mattpocock: I don't think it makes much of a visual difference, and I find the over-roundedness a bit ugly - if we do round them to 1rem we'll need to increase the padding.

davidkpiano: Alright, let's increase the left/right padding 👍

davidkpiano:
CleanShot 2021-08-24 at 09 30 38

Can we have a muted blue for inactive events? These look too similar IMO.

src/StateNodeViz.scss
```diff
@@ -64,6 +64,7 @@
 
 [data-viz='stateNode-content'] {
   background: var(--viz-node-color-bg);
+  padding: 0.6rem 1.4rem;
```
davidkpiano, line 67:
```suggestion
  padding: 0.5rem 1.5rem;
```

A little less magical 😅  (So that our "unit" can be 0.5rem)

Andarist (approved):

PR #212, opened 2021-08-24
Upgrade Next.js to 11.1.0

Andarist:


next.config.js
```diff
@@ -2,6 +2,7 @@ const registryPublicUrl =
   process.env.REGISTRY_PUBLIC_URL || `https://dev.stately.ai/registry`;
 const landingPagePublicUrl = `https://landing-page-prod.stately.ai`;
 
+/** @type import('next').NextConfig */
```
Andarist, line 5: I've decided to make this upgrade because this type was only available under some deep, nested, path in the Next.js version that we had. This has been somewhat recently changed and this type got exported from the root of the package here: https://github.com/vercel/next.js/pull/26756

davidkpiano (approved):

mattpocock (approved):

PR #214, opened 2021-08-25
Visualize multiple actors

davidkpiano:
https://user-images.githubusercontent.com/1093738/131218157-c0830c3a-4c11-4eff-ba2c-db8dffc3ecb2.mp4

Stopped actors:

CleanShot 2021-08-28 at 08 49 51@2x


src/ActorsPanel.tsx
```diff
@@ -88,21 +89,25 @@ export const ActorsPanel: React.FC = () => {
...
+            }}
+            background={
+              currentSessionId === sessionId ? 'whiteAlpha.200' : 'transparent'
+            }
+            borderRadius="md"
+            _hover={{
+              background: 'whiteAlpha.100',
+            }}
+            marginBottom="2"
+            cursor="pointer"
```
Andarist, line 105: It IMHO would make slightly more sense to wrap the whole content of the `ListItem` in `Link` - if you are using the `Link`. That way you wouldn't have add `cursor: pointer` on this element and `onClick` would be placed on a slightly more appropriate (semantically) element

src/simulationMachine.tsx
```diff
@@ -217,8 +219,13 @@ export const simulationMachine = simModel.createMachine(
...
               } else if (event.type === 'RESET') {
-                rootService?.stop();
-                locallyInterpret(rootMachine);
+                stopServices();
+                machines.forEach((machine) => {
+                  locallyInterpret(machine);
+                });
+              } else if (event.type === 'STOP') {
+                stopServices();
+                machines.clear();
```
Andarist, line 228: This could be moved into the block responsible for handling the `INTERPRET` machine - we probably want to always call this logic then, so I'm not sure if lifting this up as an additional command that can be dispatched by the parent brings much to the table. And when analyzing this part of the code having them grouped would make it slightly easier to follow.

src/Graph.tsx
```diff
@@ -315,14 +327,10 @@ function elkJSON(elkNode: StateElkNode): any {
...
 
-const GraphNode: React.FC<{ elkNode: StateElkNode }> = ({ elkNode }) => {
-  return <StateNodeViz stateNode={elkNode.node.data} node={elkNode.node} />;
-};
-
 export async function getElkGraph(
   rootDigraphNode: DirectedGraphNode,
 ): Promise<ElkNode> {
-  const rectMap = getRectMap();
+  const rectMap = await getRectMap(rootDigraphNode.id);
```
Andarist, line 333: any particular reason behind this change? I don't see it immediately when reading through those changes and it seems that it either is not needed or it has fixed some race condition - but it's hard to tell without knowing more context behind this

davidkpiano: Not a race condition; without this change, when switching between actors, `getElkGraph()` is called when the previous graph is destroyed and the new graph isn't yet rendered; therefore, `getRectMap(...)` is useless and calling `.width` was throwing an error (can't read `.width` of `undefined`).

This was introduced to fix that consistent error.

davidkpiano:
Ready for re-review

farskid:
The code below is supposed to generate 4 actors but it spawns only 3.
```js
import { createMachine, assign, spawn } from "xstate";
const machine = createMachine({
  id: "test machine",
  initial: "b",
  context: { ref: null },
  states: {
    a: {
      on: {
        GO: "b",
      },
    },
    b: {
      on: {
        GO: "a",
      },
    },
  },
});

const child = createMachine({
  id: "child",
  initial: "c",
  states: {
    c: {
      entry: assign({
        ref: () => spawn(child2),
      }),
    },
  },
});

const child2 = createMachine({
  id: "child2",
  initial: "d",
  states: {
    d: {},
  },
});
```

farskid (changes requested):
Holding this until the bug with the number of actors is fixed.

davidkpiano:
> The code below is supposed to generate 4 actors but it spawns only 3.
> 
> ```js
> import { createMachine, assign, spawn } from "xstate";
> const machine = createMachine({
>   id: "test machine",
>   initial: "b",
>   context: { ref: null },
>   states: {
>     a: {
>       on: {
>         GO: "b",
>       },
>     },
>     b: {
>       on: {
>         GO: "a",
>       },
>     },
>   },
> });
> 
> const child = createMachine({
>   id: "child",
>   initial: "c",
>   states: {
>     c: {
>       entry: assign({
>         ref: () => spawn(child2),
>       }),
>     },
>   },
> });
> 
> const child2 = createMachine({
>   id: "child2",
>   initial: "d",
>   states: {
>     d: {},
>   },
> });
> ```

The issue here is that `context` is missing:

```diff
const child = createMachine({
  id: "child",
  initial: "c",
+ context: {},
  states: {
    c: {
      entry: assign({
        ref: () => spawn(child2),
      }),
    },
  },
});
```

It works as expected with that fix ^

farskid:
> > The code below is supposed to generate 4 actors but it spawns only 3.
> > ```js
> > import { createMachine, assign, spawn } from "xstate";
> > const machine = createMachine({
> >   id: "test machine",
> >   initial: "b",
> >   context: { ref: null },
> >   states: {
> >     a: {
> >       on: {
> >         GO: "b",
> >       },
> >     },
> >     b: {
> >       on: {
> >         GO: "a",
> >       },
> >     },
> >   },
> > });
> > 
> > const child = createMachine({
> >   id: "child",
> >   initial: "c",
> >   states: {
> >     c: {
> >       entry: assign({
> >         ref: () => spawn(child2),
> >       }),
> >     },
> >   },
> > });
> > 
> > const child2 = createMachine({
> >   id: "child2",
> >   initial: "d",
> >   states: {
> >     d: {},
> >   },
> > });
> > ```
> 
> The issue here is that `context` is missing:
> 
> ```diff
> const child = createMachine({
>   id: "child",
>   initial: "c",
> + context: {},
>   states: {
>     c: {
>       entry: assign({
>         ref: () => spawn(child2),
>       }),
>     },
>   },
> });
> ```
> 
> It works as expected with that fix ^

ok. following the discussion we had on Discord, this will be the default in v5. Ship it.

farskid (approved):

PR #215, opened 2021-08-25
Added welcome area

mattpocock:
https://www.loom.com/share/e1ac5da6e7ce4009a678becf50028107﻿


.changeset/perfect-ducks-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"xstate-viz-app": patch
```
davidkpiano, line 2:
```suggestion
"xstate-viz-app": minor
```

src/EditorPanel.tsx
```diff
@@ -432,7 +432,7 @@ export const EditorPanel: React.FC<{
             {/* This extra div acts as a placeholder that is supposed to stretch while EditorWithXStateImports lazy-loads (thanks to `1fr` on the grid) */}
             <div>
               <EditorWithXStateImports
-                defaultValue={defaultValue}
+                value={value}
```
davidkpiano, line 435: Silly question, but will this cause more rerenders (since `value` is now making a round-trip with the machine)?

src/WelcomeArea.tsx
```diff
@@ -0,0 +1,178 @@
...
+                </Description>
+              </Stack>
+              <Link
+                color="blue.300"
+                fontSize="sm"
+                href="https://xstate.js.org/docs"
+                target="_blank"
+                rel="noreferrer"
+              >
+                New to XState? Read here...
```
davidkpiano, line 103:
```suggestion
                New to XState? Read the docs

File: src/sourceMachine.ts [hunk 4]
```diff
@@ -36,6 +36,24 @@ import { notifMachine, notifModel } from './notificationMachine';
...
+
+const toggleMachine = createMachine({
+  id: 'toggle',
+  initial: 'inactive',
+  states: {
+    inactive: { on: { TOGGLE: 'active' } },
+    active: { on: { TOGGLE: 'inactive' } }
+  }
+});
+`.trim();
```
davidkpiano, line 56: How about the original viz machine?

```js
const fetchMachine = createMachine({
  id: 'fetch',
  initial: 'idle',
  context: {
    retries: 0
  },
  states: {
    idle: {
      on: {
        FETCH: 'loading'
      }
    },
    loading: {
      on: {
        RESOLVE: 'success',
        REJECT: 'failure'
      }
    },
    success: {
      type: 'final'
    },
    failure: {
      on: {
        RETRY: {
          target: 'loading',
          actions: assign({
            retries: (context, event) => context.retries + 1
          })
        }
      }
    }
  }
});
```

davidkpiano (approved):
Looks great! Some nits 🔍

src/EditorPanel.tsx
mattpocock: No, I don't think so - value has been making a round trip consistently before, so I don't think that the performance drop-off is noticeable.

I thought about setting the value programmatically, but it requires lifting the editor ref up, which feels uglier to me.

src/sourceMachine.ts
mattpocock: Nice! Just need to change some text elsewhere

src/EditorPanel.tsx
Andarist: For reference - the `Editor` has to compare the old and new values to skip over forcefully executing edits:
https://github.com/suren-atoyan/monaco-react/blob/5a308001a49a10001b6d7c63dcb7be62fcf7676b/src/Editor/Editor.js#L84-L92

Even though this is a basic string comparison, the text itself can be lengthy and this will be executed on each keystroke. I've compared the responsiveness of the editor on the production and on this branch and I can feel a little bit of a slowdown, the input sometimes lags a little bit (not everyone has M1 Matt! 😉 ). However, I've also rechecked how the `dev` branch performs locally and I can see a similar slowdown there. Therefore this might be more of a React issue than those linked lines of code - sometimes the perf can be quite different between production and development builds of React.

To sum it up - I wouldn't bother with this and just observe if this comes with any production perf regressions.

.changeset/perfect-ducks-enjoy.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Added a welcome area for new users to help new folks understand what the viz can do.
```
Andarist, line 5: A screenshot of this new "screen" would look great in the release notes :P

src/Icons.tsx
```diff
@@ -87,3 +87,27 @@ export const ForkIcon = makeIcon(
     viewBox: '0 0 512 512',
   },
 );
+
+export const Lightbulb = makeIcon(
```
Andarist, line 91: are those icons taken from any external source? would be great to link to the source here, for each icon. Are there any licensing concerns related to using those icons?

Andarist (approved):

.changeset/perfect-ducks-enjoy.md
mattpocock: How, sensei?

src/Icons.tsx
mattpocock: Yes, Material! No licensing concerns

mattpocock: https://github.com/google/material-design-icons/blob/master/LICENSE

.changeset/perfect-ducks-enjoy.md
Andarist: 1. upload to any location - u can even upload to GitHub by dropping a screenshot into the comment box and once the upload is done just copy-paste the result
2. use `` tag in the changeset file 😉

src/Icons.tsx
Andarist: According to this text, we should include a copy of their LICENSE somewhere - but quite frankly, I never know how to do it properly.

PR #226, opened 2021-08-29
Collapsible panels

davidkpiano:
This PR adds support for collapsing/expanding panels, which also means the header needed to be refactored

https://user-images.githubusercontent.com/1093738/131265752-29192814-a218-43e2-8531-9699adeab704.mp4

TODO:

- [x] Fix the global interpreted machine


mattpocock:
image

mattpocock:
Don't love the top panel - detailed feedback here:

https://www.loom.com/share/4668c64fc5f9433ca7d58d61286fd3e4

farskid:
Strange transition on resizable box
https://www.loom.com/share/9cf15abcca3b45bab3fdd6ae9b2a8e9f

farskid:
Also look at the line numbers of the editor. They hide immediately after the first load and while scrolling in the editor, the tab panels hide behind the top bar.
https://www.loom.com/share/c427a2316a0647caba92413735ba4586

Andarist:
Something quite weird is going on with tabbing - basically, the panels on the right can be focused this way when using the keyboard navigation (I assume that it's because the elements there stay in the document and they are not removed from the tab sequence as the container doesn't have `display: none` or anything like that). They can, in a similar fashion, disappear when you tab out of the container - but if you click somewhere else the panel can stay visible. Note that, in this situation, this doesn't match the "hide/show" button correctly.

My suspicion is that this is browser forcefully scrolling to the focused element - but whatever the reason is: those elements there should not be focusable when panels are hidden.

On this video, I **only** press Tab (and Alt+Tab to tab out of the Monaco editor): https://www.loom.com/share/b2264fd05faa4a838fd32fe2997e2671

This can also lead to weird/broken layouts like this:
Screenshot 2021-08-30 at 17 50 19

src/CanvasView.tsx
```diff
@@ -144,42 +141,50 @@ export const CanvasView: React.FC = () => {
...
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  Privacy Policy
+                </MenuItem>
+              </MenuList>
+            </Portal>
+          </Menu>
+          <IconButton
+            aria-label="Collapse panels"
```
Andarist, line 181: This should be conditional, like the `icon` prop, as well. The aria-label should describe the currently available action accurately.

src/CanvasView.tsx
```diff
@@ -144,42 +141,50 @@ export const CanvasView: React.FC = () => {
...
+                />
+              }
+            />
+            <Portal>
+              <MenuList fontSize="sm" padding="0">
+                <MenuItem
+                  as={Link}
+                  href="https://github.com/statelyai/xstate"
+                  target="_blank"
+                  rel="noreferrer"
```
Andarist, line 165:
```suggestion
                  rel="noopener noreferrer"
```

src/CanvasView.tsx
```diff
@@ -144,42 +141,50 @@ export const CanvasView: React.FC = () => {
...
+                >
+                  XState version 4.23.0
+                </MenuItem>
+                <MenuItem
+                  as={Link}
+                  href="https://stately.ai/privacy"
+                  target="_blank"
+                  rel="noreferrer"
```
Andarist, line 173:
```suggestion
                  rel="noopener noreferrer"
```

src/HeaderView.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <Link
+          href="/"
+          title="Stately.ai"
+          display="block"
+          height="100%"
+          _hover={{
+            opacity: 0.8,
+          }}
+          target="_blank"
+          rel="noreferrer"
```
Andarist, line 44:
```suggestion
          rel="noopener noreferrer"
```

src/HeaderView.tsx
```diff
@@ -0,0 +1,114 @@
...
+                <MenuList>
+                  {userOwnsSource && sourceState.context.sourceID && (
+                    <MenuItem
+                      as="a"
+                      href={registryLinks.editSourceFile(
+                        sourceState.context.sourceID,
+                      )}
+                    >
+                      <HStack spacing="3">
+                        <EditIcon />
```
Andarist, line 83:
```suggestion
                        <EditIcon aria-hidden />
```

src/HeaderView.tsx
```diff
@@ -0,0 +1,114 @@
...
+                      </HStack>
+                    </MenuItem>
+                  )}
+                  {registryData.owner && (
+                    <MenuItem
+                      as="a"
+                      href={registryLinks.viewUserById(registryData?.owner?.id)}
+                    >
+                      <HStack spacing="3">
+                        <Avatar
```
Andarist, line 94: I'm not entirely sure - but it feels that it might be preferable here to treat this similar to other decorative icons in this Menu. So maybe using `aria-hidden` on this one would be appropriate?

src/CanvasView.tsx
```diff
@@ -144,42 +141,50 @@ export const CanvasView: React.FC = () => {
...
-                href="https://stately.ai/privacy"
-                target="_blank"
-                rel="noreferrer"
-              >
-                Privacy Policy
-              </MenuItem>
-            </MenuList>
-          </Portal>
-        </Menu>
+        <ButtonGroup size="sm" spacing={2} marginLeft="auto">
```
Andarist, line 144: This button group doesn't move together with the panel when it's getting collapsed/expanded - which is IMHO suboptimal since you can see it jumping from one spot to another when the other element is animating.

src/PanelsView.tsx
```diff
@@ -22,26 +13,42 @@ import { SpinnerWithText } from './SpinnerWithText';
...
         height="100%"
       >
-        <TabList>
+        <TabList textTransform="uppercase">
           <Tab>Code</Tab>
           <Tab>State</Tab>
           <Tab>Events</Tab>
           <Tab>Actors</Tab>
           <Tab marginLeft="auto" marginRight="2">
             <SettingsIcon />
```
Andarist, line 50:
```suggestion
            <SettingsIcon aria-label="Settings" />
```

src/CanvasHeader.tsx
```diff
@@ -21,10 +21,10 @@ import { useSourceActor } from './sourceMachine';
 
 export const CanvasHeader: React.FC = () => {
```
Andarist, line 22: This file should now be removed, right? It doesn't seem to be used anywhere

src/HeaderView.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        <Text>View Author</Text>
+                      </HStack>
+                    </MenuItem>
+                  )}
+                </MenuList>
+              </Menu>
+            </HStack>
+          </Stack>
+        )}
+        <Login />
```
Andarist, line 110: The avatar in this component isn't aligned with the SettingsIcon

src/CanvasView.tsx
davidkpiano: Not a blocker though, right? Something we can take care of in a separate PR.

Moving it with the panel is non-trivial.

Andarist: Yes, not a blocker. To make the implementation easy easier we could try to level up our Framer Motion game.

davidkpiano:
@mattpocock @farskid @Andarist Feedback addressed:

https://user-images.githubusercontent.com/1093738/131406675-f96ecce7-5e59-433e-b5ad-c82c2e618c24.mp4

mattpocock:
Yes, this is better. The login button no longer appears to have `colorScheme="blue"` on it - it would be good if this came back

mattpocock (approved):
Yarp, happy with this now

src/HeaderView.tsx
davidkpiano: This wasn't modified

src/HeaderView.tsx
davidkpiano: They're no longer adjacent

farskid:
@davidkpiano Before merging, let's update this to the latest dev branch and check it in embedded mode.

cypress/integration/visual.spec.ts
```diff
@@ -62,4 +62,30 @@ describe('Visual', () => {
...
   });
+
+  it('show/collapsing panels button should work', () => {
+    cy.visit('/viz');
+
+    cy.findByTestId('panels').then(($panels) => {
+      expect(isInViewport($panels)).eq(true);
+    });
+
+    cy.findByRole('button', { name: 'Collapse panels' }).click();
```
farskid, line 73: nit: I find regex safer in targetting elements with names.
```suggestion
    cy.findByRole('button', { name: /collapse panels/i }).click();
```

cypress/integration/visual.spec.ts
```diff
@@ -62,4 +62,30 @@ describe('Visual', () => {
...
+      expect(isInViewport($panels)).eq(false);
+    });
+
+    cy.findByRole('button', { name: 'Show panels' }).click();
+
+    cy.findByTestId('panels').then(($panels) => {
+      expect(isInViewport($panels)).eq(false);
+    });
+
+    cy.findByRole('button', { name: 'Visualize' }).then(($button) => {
```
farskid, line 85: nit:
```suggestion
    cy.findByRole('button', { name: /visualize/i }).then(($button) => {
```

farskid: a nit comment for all the queries.

farskid:
image
There is a layering problem as shown above. The canvas should be below the header all the time.

farskid:
When the right panels are closed, the icon still remains the same.
image

PR #228, opened 2021-08-30
Fix getLCA

### LINKED ISSUE #227: weird, less readable location of some transitions in new inspector vs. old
As [mentioned in Discord recently](https://discord.com/channels/795785288994652170/873201691032420442/880909005394956299)

[Code sandbox here](https://codesandbox.io/s/pedantic-snow-omqke) with both inspector URLs to comment in and out.

And here's the screenshots posted in Discord:
old viz
new viz

davidkpiano:
This PR fixes the `getLCA(...)` function. The problem was that it was previously not considering the source node to potentially be a LCA.

Fixes #227 

**Before:**
Before

**After:**
CleanShot 2021-08-29 at 21 53 00

### DIFF
--- a/.changeset/tall-boats-melt.md
+++ b/.changeset/tall-boats-melt.md
@@ -3,3 +3,5 @@
 ---
 
 The layout algorithm was parenting some of the edges incorrectly. This has been fixed.
+
+![CleanShot 2021-08-30 at 07 14 58](https://user-images.githubusercontent.com/1093738/131331546-4a780c96-c58e-498e-aeaa-f44dc5e81fb4.png)
--- a/.changeset/tall-boats-melt.md
+++ b/.changeset/tall-boats-melt.md
@@ -1,0 +1,5 @@
+---
+'xstate-viz-app': patch
+---
+
+The layout algorithm was parenting some of the edges incorrectly. This has been fixed.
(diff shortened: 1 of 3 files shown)


Andarist:
> The problem was that it was previously not considering the source node to potentially be a LCA.

That's because the source node cannot be LCA - it's in the name: Lowest Common **Ancestor**. A node itself cannot be its own ancestor. 

I'm not questioning the changes though - because I don't understand them yet. However, If those changes are required/correct then the containing function should be renamed.

Could you go into more detail about what was happening and how it got fixed? From the screenshot it's not really obvious - this is how this is being rendered on production, right? Could you also depict how this is rendered after your changes? Attaching a source code for this machine would also be helpful in understanding what this machine is doing and which part of it gets rendered in a confusing manner.

davidkpiano:
> > The problem was that it was previously not considering the source node to potentially be a LCA.
> 
> That's because the source node cannot be LCA - it's in the name: Lowest Common **Ancestor**. A node itself cannot be its own ancestor.
> 
> I'm not questioning the changes though - because I don't understand them yet. However, If those changes are required/correct then the containing function should be renamed.
> 
> Could you go into more detail about what was happening and how it got fixed? From the screenshot it's not really obvious - this is how this is being rendered on production, right? Could you also depict how this is rendered after your changes? Attaching a source code for this machine would also be helpful in understanding what this machine is doing and which part of it gets rendered in a confusing manner.

Did you see the linked issue? https://github.com/statelyai/xstate-viz/issues/227

Andarist (approved):
Could you add a changeset for this improvement?

mattpocock (approved):

Issue #225, opened 2021-08-29
Horizontal scroll not working on Windows


a-uz:
It seems that pressing shift key while scrolling the mouse wheel has no effect on Windows, contents of the canvas move up or down, instead of left or right. On Mac it works as expected.

I was able to reproduce this with the latest versions of Chrome and Edge on Windows 10.

Andarist:
Thank you for the report. I've opened a draft PR with the fix here: https://github.com/statelyai/xstate-viz/pull/230

a-uz:
Just tried the fix and it works great!

Andarist:
@a-uz could you confirm that you are using an actual mouse? I have some doubts about this fix: https://github.com/statelyai/xstate-viz/pull/230#discussion_r698792935

a-uz:
@Andarist Yes, I'm using an actual mouse.

I did some more testing and with the current production version on Windows trackpad interactions work just fine, but shift key + mouse wheel doesn't. With the PR version shift key + mouse wheel works, but shift key + trackpad is now strange: I can only drag up and down to pan left or right and there's no way to pan up or down. TBH, I'm not even sure how it should be.

a-uz:
I tried to use shift key + trackpad on Figma and Draw.io. On Figma's canvas holding shift key and dragging up or down pans left or right, dragging left or right pans up or down. On Draw.io's canvas holding shift key and dragging in all directions pans left or right.

Andarist:
@a-uz I've pushed out some changes, could you give this a spin? I would also appreciate it if you could rephrase your Figma & Draw.io research because I'm unsure if I got this right. Mentioning the exact combination of shift, wheel/trackpad, physical direction and the resulting pan direction for all of those cases would help me a lot.

a-uz:
@Andarist Now the mouse + shift key combo doesn't work anymore and trackpad + shift key works the same as without the shift key.

Dragging on Figma's canvas with trackpad + shift key inverts the direction: dragging vertically pans horizontally and dragging horizontally pans vertically. Dragging on Draw.io's canvas with trackpad + shift key pans horizontally regardless of drag direction. 

I guess I would prefer the visualizer to have Figma's approach by inverting the direction when shift key is pressed.

Andarist:
Ok, thank you for your feedback - gonna take another look at this tomorrow. I'm a little bit surprised though that shift+mouse doesn't work - maybe some kind of a silly mistake in the adjustments I've made.

[...]

PR #230, opened 2021-08-30
Fixed an issue with horizontal scrolling on Windows not moving canvas correctly

Andarist:
fixes https://github.com/statelyai/xstate-viz/issues/225

### DIFF
--- a/src/CanvasContainer.tsx
+++ b/src/CanvasContainer.tsx
@@ -4,7 +4,7 @@ import { useCanvas } from './CanvasContext';
 import { useMachine } from '@xstate/react';
 import { createModel } from 'xstate/lib/model';
 import { Point } from './pathUtils';
-import { isWithPlatformMetaKey, isTextInputLikeElement } from './utils';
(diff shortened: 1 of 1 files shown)


src/CanvasContainer.tsx
```diff
@@ -204,7 +204,11 @@ export const CanvasContainer: React.FC = ({ children }) => {
           );
         }
       } else if (!e.metaKey && !e.ctrlKey) {
-        canvasService.send(canvasModel.events.PAN(e.deltaX, e.deltaY));
+        if (isMac() || !e.shiftKey) {
+          canvasService.send(canvasModel.events.PAN(e.deltaX, e.deltaY));
+        } else {
+          canvasService.send(canvasModel.events.PAN(e.deltaY, 0));
```
Andarist, line 210: After playing with this a little bit more I'm unsure if this is the best fix for the issue. It works for this specific case but I'm hesitant to say that it's universally good. It would be quite good if I could only apply this for mouse interactions but as far as I can tell they are indistinguishable from touchpad interactions.

PR #235, opened 2021-08-31
Use promisified `fs.readFile` instead of `fs/promises` since that module is not available in node12

Andarist:


davidkpiano:
Do we have to use Node 12? Or can we use the latest version?

Andarist:
> Do we have to use Node 12? Or can we use the latest version?

For the time being it's a requirement - at least for default images on the GitHub Actions runner. They will be updating that but it's uncertain when that will be (most likely in this year though).

mattpocock (approved):

davidkpiano (approved):

PR #238, opened 2021-08-31
Make all machines fit to view by default

mattpocock:


davidkpiano (approved):

davidkpiano (approved):

src/canvasMachine.tsx
```diff
@@ -186,32 +186,39 @@ export const canvasMachine = canvasModel.createMachine({
...
     },
     'elkGraph.UPDATE': {
-      actions: canvasModel.assign({
-        elkGraph: (_, e) => e.elkGraph,
-      }),
+      actions: [
+        canvasModel.assign({
+          elkGraph: (_, e) => e.elkGraph,
+        }),
+        send('FIT_TO_VIEW'),
```
farskid, line 193: Does this trigger when the source updates through `inspect` mode?

src/canvasMachine.tsx
```diff
@@ -186,32 +186,39 @@ export const canvasMachine = canvasModel.createMachine({
...
+      ],
     },
     FIT_TO_VIEW: {
       actions: [
         canvasModel.assign({
           zoom: (ctx) => {
+            if (!ctx.elkGraph) return ctx.zoom;
```
farskid, line 200: why do we need this?

mattpocock: Because `ctx.elkGraph` can be undefined, and if it is we need to do nothing, silently

src/canvasMachine.tsx
mattpocock: Yes - whenever the elk graph calculation updates, this will update.

PR #240, opened 2021-08-31
Support timers in realm

### LINKED ISSUE #205: setTimeout & setInterval not working
Problem:
**setTimeout** & **setInterval** can't be used on the visualizer right now. It gives the error: "setTimeout is not a function".

Demo: https://stately.ai/viz?id=a52dd233-64c9-4476-a869-cd6b58b10a12#

mattpocock:
Fixes #205

### DIFF
--- a/src/parseMachine.ts
+++ b/src/parseMachine.ts
@@ -15,16 +15,16 @@ const wrapCallbackToPreventThis = (callback: () => void) => () => {
 };
 
 const windowShim = {
-  setInterval: function (callback: () => void, ...args: any[]) {
+  setInterval: (callback: () => void, ...args: any[]) => {
     return setInterval(wrapCallbackToPreventThis(callback), ...args);
   },
-  setTimeout: function (callback: () => void, ...args: any[]) {
+  setTimeout: (callback: () => void, ...args: any[]) => {
     return setTimeout(wrapCallbackToPreventThis(callback), ...args);
   },
-  clearTimeout: function (...args: any[]) {
+  clearTimeout: (...args: any[]) => {
     return clearTimeout(...args);
   },
-  clearInterval: function (...args: any[]) {
+  clearInterval: (...args: any[]) => {
     return clearInterval(...args);
   },
 };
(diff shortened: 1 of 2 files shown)


src/parseMachine.ts
```diff
@@ -54,6 +54,10 @@ export function parseMachines(sourceJs: string): Array<StateNode> {
       log: console.log,
       warn: console.warn,
     },
+    setInterval,
+    setTimeout,
+    clearTimeout,
+    clearInterval,
```
Andarist, line 60: Have you tested this locally though? :troll:

mattpocock: Ooo, good point

mattpocock: @Andarist Ready for re-review!

src/parseMachine.ts
```diff
@@ -6,6 +6,29 @@ import realmsShim from 'realms-shim';
 
 const realm = realmsShim.makeRootRealm();
 
+/**
+ * Wraps a callback in a function to prevent
+ * accessing 'this'
+ */
```
Andarist, line 12: This basically repeats the name of the function - do we need comments like this? 🤔

src/parseMachine.ts
```diff
@@ -6,6 +6,29 @@ import realmsShim from 'realms-shim';
...
+const wrapCallbackToPreventThis = (callback: () => void) => () => {
+  return callback();
+};
```
Andarist, line 15: Something like this should improve the compatibility with native timers:
```suggestion
const wrapCallbackToPreventThis = (callback: (...args: any[]) => void) => (...args: any[]) => {
  return callback(...args);
};
```

`setTimeout` and `setInterval` forward arguments to the inner callback, so this works:
```js
setTimeout((a, b) => {}, 0, 'a', 'b')
```

src/parseMachine.ts
mattpocock: Sure, I can ditch if you like

src/parseMachine.ts
mattpocock: Mmm, good catch

davidkpiano (approved):

Andarist (approved):

karlguillotte:
Are you expecting to deploy that fix to the live Viz soon? I am prototyping a machine with the Viz and "setTimeout" is a useful function to have for testing. 

Thanks!

PR #246, opened 2021-09-01
Andarist/sta 560 hand tool

Andarist:
This PR introduces a "pan mode" that can be enabled by pressing a hand icon in our canvas controls toolbar. This allows one to pan around the canvas without having to press Space constantly which is an accessibility feature. It might be hard to both hold Space down and operate a mouse at the same time.

This is how it looks in the toolbar:
Screenshot 2021-09-13 at 10 49 14

And this is how it looks pressed:
Screenshot 2021-09-13 at 10 49 28

The core of the changes is just in the `CanvasContainer` file. I've refactored the machine there completely while also extracting "drag session tracker" to a separate machine that gets invoked by the parent. One thing worth noting is that I've used coordinated parallel regions pattern in the parent machine. While this is based on kinda an implicit relation between those regions (one raises an event and the second one transitions based on it) it has allowed me to keep the overall modeling of this pretty simple and reuse a panning region for both the classic panning with a space and for the pan/hand mode. I'm very curious to hear out what do you think about it.

For comparison, I put the before & after of the machine(s) here.

**Before**
Screenshot 2021-09-12 at 10 24 00

**After**
Screenshot 2021-09-12 at 15 49 19

Screenshot 2021-09-12 at 17 27 50

Some minor implementation changes:
- moved attaching pointer listeners to a machine - this allows us to neatly avoid checking if an event should be dispatched to a machine or not in the React event listeners. This requires passing `ref` to a machine but this seems to be mostly OK even if not something that we do anywhere else. I feel like we could leverage this pattern more, wdyt?
- `PAN` gets sent to the `canvasService` from within an action - instead of being synchronized through effect
- removed disabling/enabling of the text selection. I've tried hard (experimented with different interactions with that and without that) but I couldn't see what it has been fixing. It might be that it was needed at some stage but is no longer needed. I could also just be wrong and maybe I've missed what exactly it has been fixing. If you find anything broken because of that it will be very easy to bring it back again - I would appreciate you testing this out
- ensured that only a single (and the same) pointer is handled throughout the panning session
- added `onpointercancel` handling
- ensured that the panning happens only with the left button the mouse

Bug fixes:
- fixed an issue with buttons in the application not being "clickable" with Space (previously `invokeDetectLock` was calling `ev.preventDefault()` for buttons)

fixes STA-258


src/Icons.tsx
```diff
@@ -15,35 +15,35 @@ const makeIcon =
     );
   };
 
-export const ThumbUpOutlined = makeIcon(
+export const ThumbUpOutlinedIcon = makeIcon(
```
Andarist, line 18: Just unified naming scheme of the exported icons

src/base.scss
```diff
@@ -17,7 +17,6 @@
   }
 
   --viz-color-transparent: #fff6;
-  --viz-color-active: turquoise;
```
Andarist, line 20: to the best of my knowledge this was unused

src/CanvasContainer.tsx
```diff
@@ -2,171 +2,400 @@ import React, { CSSProperties, useEffect, useRef } from 'react';
...
-          on: {
-            GRAB: {
-              target: 'grabbed',
-              actions: dragModel.assign({
-                startPoint: (_, e) => e.point,
-                dragPoint: (_, e) => e.point,
-              }),
+        },
+      },
+      active: {
```
mattpocock, line 82: We should add the functionality where it disables text selection when active, see the old changes in this file. Adding them on the entry/exit here would make sense?

Andarist: from the description of the PR:
> removed disabling/enabling of the text selection. I've tried hard (experimented with different interactions with that and without that) but I couldn't see what it has been fixing. It might be that it was needed at some stage but is no longer needed. I could also just be wrong and maybe I've missed what exactly it has been fixing. If you find anything broken because of that it will be very easy to bring it back again - I would appreciate you testing this out

mattpocock (approved):
This looks good to me, no regressions. Have added a suggestion for the text selection.

src/CanvasContainer.tsx
mattpocock: Gotcha - yes, I did find that text selection was flickering when I was dragging in certain spots.

mattpocock: Want a loom? You can just copy the previous implementation to add it back in.

Andarist: Ye - I would love a loom of this to understand this interaction better. I will re-add this before merging this PR.

src/CanvasContainer.tsx
```diff
@@ -2,171 +2,400 @@ import React, { CSSProperties, useEffect, useRef } from 'react';
...
+
+const dragSessionModel = createModel(
   {
-    startPoint: { x: 0, y: 0 },
-    dragPoint: { x: 0, y: 0 },
-    dx: 0,
-    dy: 0,
-    embed: undefined as EmbedContext | undefined,
+    session: null as DragSession | null,
+    ref: null as React.MutableRefObject<HTMLElement> | null,
```
farskid, line 31: We might need a better naming to what this ref refers.

farskid: Also might be better to just save the HTML element in the context instead of tightening the model to React.

src/CanvasContainer.tsx
```diff
@@ -2,171 +2,400 @@ import React, { CSSProperties, useEffect, useRef } from 'react';
...
+    states: {
+      idle: {
+        invoke: {
+          id: 'dragSessionStartedListener',
+          src:
+            ({ ref }) =>
+            (sendBack) => {
+              const node = ref!.current!;
+              const listener = (ev: PointerEvent) => {
+                const isLeftButton = ev.button === 0;
```
farskid, line 58:
```suggestion
                const isLeftMouseButton = ev.button === 0;
```

farskid: is this check against `ev.buttton` consistent for other types of input other than a physical mouse?

src/CanvasContainer.tsx
```diff
@@ -2,171 +2,400 @@ import React, { CSSProperties, useEffect, useRef } from 'react';
...
-        src: 'invokeDetectRelease',
-      },
-      states: {
-        idle: {
-          meta: {
-            cursor: 'grab',
+        on: {
+          DRAG_SESSION_STARTED: {
+            target: 'active',
+            actions: actions.forwardTo(SpecialTargets.Parent),
```
farskid, line 78: This is the first time I see `SpecialTargets` from XState. is this the same as `sendParent` the event?

src/CanvasContainer.tsx
```diff
@@ -198,22 +427,11 @@ export const CanvasContainer: React.FC = ({ children }) => {
    */
   useEffect(() => {
     const onCanvasWheel = (e: WheelEvent) => {
-      const isWithMetaKey = isWithPlatformMetaKey(e);
+      const isZoomEnabled = !embed?.isEmbedded || embed.zoom;
+      const isPanEnabled = !embed?.isEmbedded || embed.pan;
```
farskid, line 431: isn't it easier to think of these 2 conditions from the negative side?
```js
const isZoomDisabled = embed?.isEmbedded && !embed.zoom;
```
just reads better than a positive check.

farskid (changes requested):
Left a few comments. Let me know if you think they're nit.

farskid:
Fixes https://linear.app/statelyai/issue/STA-258/refactor-drag-machine-to-encapsulate-all-the-logic-scattered-through

farskid:
> Fixes https://linear.app/statelyai/issue/STA-258/refactor-drag-machine-to-encapsulate-all-the-logic-scattered-through

@mattpocock Does this comment get picked up by Linear or should it be in the PR description?

farskid:
@Andarist Here is a loom showing the machine ID text is selected while dragging
https://www.loom.com/share/d005abf1556a40acbbf460d670395300

farskid:
@Andarist Although pan mode and panning by Spacebar are 2 different features, they're still the same to the end-user and I think pan mode should be enabled while panning by Spacebar.

src/CanvasContainer.tsx
Andarist: Yes, it's just that I've wanted to use that with `forwardTo`

src/CanvasContainer.tsx
Andarist: It is consistent in terms of the desired action - I think we should always assume that those are intentions to drag (for touch and pen this means "contact": https://w3c.github.io/pointerevents/#the-button-property ).

The variable might be named slightly inaccurate though - do you think I should rename it? How should I name it? `isLeftMouseButtonOrContact`? Or maybe I should just inline the condition to avoid naming stuff 😅

src/CanvasContainer.tsx
Andarist: > We might need a better naming to what this ref refers to.

From the point of this machine, it's an "abstract" ref - it doesn't refer to any particular in-app ref as the machine is generalized (I think it could actually be reused with success and could replace the one defined in the [ResizableBox](https://github.com/statelyai/xstate-viz/blob/c36ce2329c8bf0a900134be03733b27ad3502f60/src/ResizableBox.tsx#L23)).

The `ref` usually refers to a DOM ref (if it's not using some other name) so I've assumed that using this here would be OK. Would you prefer something like `domRef`?

> Also might be better to just save the HTML element in the context instead of tightening the model to React.

It could but I would have to sync the ref to a machine when the component mounts which seemed like a hassle and something that would complicate the implementation here.

src/CanvasContainer.tsx
Andarist: I just prefer to use a single flavor across the app - especially once you start creating utils/selectors for this kind of stuff and also to avoid double negation with stuff like `!isZoomDisabled`. "not not" seems slightly harder to read at times (it isn't particularly hard to read here though). 

I don't mind changing it - let me know what do you think

Andarist:
> @mattpocock Does this comment get picked up by Linear or should it be in the PR description?

Unsure but once I've added it to the PR description - it clicked. It might be the poster issue though (I might be treated differently since I'm the author of the PR). Either way - I've linked this to that Linear issue.

> @Andarist Although pan mode and panning by Spacebar are 2 different features, they're still the same to the end-user and I think pan mode should be enabled while panning by Spacebar.

Do you mean that the icon in the toolbar should look "pressed" when I pan with a Spacebar? It would be kinda confusing for me and I wouldn't think of this myself but I see that it works like that in Whimsical 🤔

src/CanvasContainer.tsx
farskid: >It could but I would have to sync the ref to a machine when the component mounts which seemed like a hassle and something that would complicate the implementation here.
I mean just passing `ref.current` directly to the machine. we won't need to sync anything back from the machine to the component so ref in the machine is always read-only.

Andarist: @farskid that's impossible - we pass `ref` to a machine during render but `ref.current` gets assigned for the first time in the commit phase of React.

farskid: ok got it.

src/CanvasContainer.tsx
farskid: Hard to find a name that satisfies all the inputs. Maybe it's fine to only name it for the mouse and move on. FWIW, `isLeftButton` is not readable to me.

src/CanvasContainer.tsx
farskid: I don't have a strong opinion over this. I'm in the minority who prefer negative variable names for toggle states.

farskid:
>Do you mean that the icon in the toolbar should look "pressed" when I pan with a Spacebar? It would be kinda confusing for me and I wouldn't think of this myself but I see that it works like that in Whimsical 🤔

Yes. To the end-user, they're both the same thing, Pan the canvas. So it makes sense to align the two.

src/CanvasContainer.tsx
Andarist: The loom from Matt: https://www.loom.com/share/9a14e0fcdb1f4324b973ba2727af897d

I still can't quite repro it myself but this means that those lines related to disabling & enabling text selection have to come back. Gonna do that in a sec

farskid (approved):

PR #245, opened 2021-09-01
Andarist/fix production saving crash

Andarist:

PR #247, opened 2021-09-01
Fixed an issue with the whole app being remounted after saving a source

Andarist:
The fixed issue is described in a comment here:
https://github.com/statelyai/xstate-viz/blob/c8c144bddeb7b9731902182546b413f774660d4a/src/pages/index.tsx#L22-L24

And the issue is not just about performance - when the app was being remounted the Monaco editor was going all white, images were remounted so a glitch could be observed, etc.

I think I would still recommend using an alternative fix - which is to merge those 2 routes and treat `[sourceFileId]` parameter as optional. It seems that it's possible to mimic that using [optional catch all pattern](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes). That can be a refactor item though.

--- 

EDIT:

after recognizing some more issues I've went for consolidating those 2 paths, using the mentioned optional catch all pattern. 

fixes STA-580


src/App.tsx
```diff
@@ -44,32 +43,35 @@ function App() {
     sourceID,
   });
 
-  if (!isOnClientSide()) return null;
+  if (!isOnClientSide()) return <AppHead {...headProps} />;
```
Andarist, line 46: This was previously rendered by route components but since those have to be the same to avoid app remounting I had to move rendering `AppHead` to this component

src/App.tsx
```diff
@@ -44,32 +43,35 @@ function App() {
...
-              </CanvasProvider>
-              <PanelsView />
-              <MachineNameChooserModal />
-            </Box>
-          </SimulationProvider>
-        </PaletteProvider>
-      </EditorThemeProvider>
-    </ChakraProvider>
+    <>
+      <AppHead {...headProps} />
```
Andarist, line 50: Just wrapped in a fragment and added `<AppHead />`

src/AppHead.tsx
```diff
@@ -19,10 +20,10 @@ interface AppHeadProps {
   importElk: boolean;
   description: string;
   importPrettier: boolean;
-  ogImageUrl: string | undefined;
+  ogImageUrl: string | null;
```
Andarist, line 23: changed to null because now this is sent in the SSRed props and undefined is not JSON-serializable

src/pages/_app.tsx
```diff
@@ -33,34 +33,10 @@ if (
...
-     * Apologies for this line of code. The reason this is here
-     * is that XState + React Fast Refresh causes an error:
-     *
-     * Error: Unable to send event to child 'ctx => ctx.sourceRef'
-     * from service 'auth'.
-     *
-     * router.replace causes this in development, but not in prod
-     *
-     * So, we use window.location.href in development (with the /viz
-     * prefix which Next won't automatically add) and router.replace in prod
```
Andarist, line 47: After this refactor I couldn't find any problems during development and also before my latest fix ( https://github.com/statelyai/xstate-viz/pull/245 ) this was an issue in production. The state of the world/codebase could have changed in the meantime and I think this is no longer necessary.

src/sourceMachine.ts
```diff
@@ -610,7 +610,9 @@ export const makeSourceMachine = (params: {
           };
         }),
         updateURLWithMachineID: (ctx) => {
-          params.routerReplace(`/${ctx.sourceID}`);
+          params.router.replace(`/${ctx.sourceID}`, undefined, {
+            shallow: true,
```
Andarist, line 614: using shallow like this won't update meta tags in the head - it seems though like fixing this could be moved to an issue

src/pages/_app.tsx
mattpocock: Very nice!

src/sourceMachine.ts
mattpocock:
```suggestion
            // using shallow like this won't update meta tags in the head
            shallow: true,
```

Andarist:
> When you save a file for the first time (in other words, after a router.replace), the Monaco editor becomes unusable. It loses all numbers down the left side, the cursor no longer shows - all sorts of weird stuff happening.

Hm, that's probably because `AppHead` rerenders - but I thought that I've fixed that. Thanks for spotting - gonna look into this later

src/pages/index.tsx
```diff
@@ -1,21 +1,23 @@
...
-  );
-};
+export const getStaticProps: GetStaticProps<{ headProps: AppHeadProps }> =
+  async () => {
+    return {
+      props: {
+        headProps: {
+          title: 'XState Visualizer',
+          ogTitle: 'XState Visualizer',
+          description: 'Visualizer for XState state machines and statecharts',
```
davidkpiano, line 12: Side-note: should we call this `Stately VIz` now? I can go either way.

tsconfig.json
```diff
@@ -18,14 +14,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "preserve",
-    "strictNullChecks": true
```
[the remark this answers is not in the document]
davidkpiano: Is `strict: true` by default?

src/pages/index.tsx
farskid: Sounds about right.

src/types.ts
```diff
@@ -23,6 +24,12 @@ export type AnyState = State<any, any>;
 
 export type SourceProvider = 'gist' | 'registry';
 
+export interface SourceRegistryData extends SourceFileFragment {
+  // we can't trust SSR data to be accurate because at the moment we can use authenticated user during SSR
```
mattpocock, line 28: Why can't we use authenticated user info during SSR? If we have access to the cookie? Or do you mean that we don't.

Andarist: We don't have access to the cookie because Supabase keeps the token in the `localStorage`. I had a cookie-based version (using some hacks~) locally but that would mean that we would be keeping the refresh token in more than one place (security concern) or that we wouldn't be able to refresh tokens during SSR (back to square one - some requests couldn't be reliably SSRed).

src/sourceMachine.ts
Andarist: With yesterday changes I've solved updating meta tags in the head

tsconfig.json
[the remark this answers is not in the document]
Andarist: it is not - but we have it enabled

src/types.ts
Andarist: btw. I have not ended up using the ssr/client distinction - at first I thought I've though that I will need it but it turned out not to be true. I've decided to keep this property though as it seems to have some informative value

cypress/integration/embedded-mode.spec.ts
```diff
@@ -1,20 +1,6 @@
...
-};
-
-describe('Embedded mode', () => {
-  describe('default (mode:viz)', () => {
-    before(() => {
-      cy.interceptGraphQL({
-        getSourceFile: {
-          id: SOURCE_ID,
-          text: `
+const sourceFileFixture = {
```
Andarist, line 1: no functional changes in this file - I've just created typed `cy.visitEmbedWithNextPageProps` and used it here (since I had to adjust this code anyway to cover for `GetSourceFileSsrQuery` being removed

package.json
```diff
@@ -11,7 +11,7 @@
     "@monaco-editor/react": "^4.2.1",
     "@sentry/react": "^6.10.0",
     "@sentry/tracing": "^6.10.0",
-    "@supabase/supabase-js": "^1.15.1",
+    "@supabase/supabase-js": "^1.22.5",
```
Andarist, line 14: comes with a fix for accessing `localStorage` during initialization of the module which has been throwing in incognito mode (and probably under some privacy settings as well). Thanks for the upgrade I could finally remove our patch for this issue

src/App.tsx
```diff
@@ -2,24 +2,52 @@ import { Box, ChakraProvider } from '@chakra-ui/react';
...
+const defaultHeadProps = {
+  title: 'XState Visualizer',
+  ogTitle: 'XState Visualizer',
+  description: 'Visualizer for XState state machines and statecharts',
+  // TODO - get an OG image for the home page
+  ogImageUrl: null,
+};
+
+const VizHead = () => {
+  const sourceRegistryData = useSourceRegistryData();
```
Andarist, line 34: This allows us to keep `<head/>` in sync with in-app data despite the app not being remounted (previously it was in sync because the whole app was being remounted)

src/authMachine.ts
```diff
@@ -29,19 +22,25 @@ import {
...
 import { storage } from './localCache';
 import { gQuery } from './utils';
-import { GetSourceFileSsrQuery } from './graphql/GetSourceFileSSR.generated';
+import { SourceRegistryData } from './types';
+import { NextRouter } from 'next/router';
 
 const authModel = createModel(
   {
-    client: null! as SupabaseClient,
+    client: createClient(
```
Andarist, line 30: Moving it here fixes andarist/sta-580-refactor-auth-machine-to-prevent which made our app spam the Supabase backend with requests when refreshing the token and when getting the user after oauth redirect

src/cookies.ts
```diff
@@ -0,0 +1,38 @@
+export function get(key: string): string | null {
```
Andarist, line 1: I didn't end up using it but removing this module seemed like a waste - might come handy in the future.

src/monacoPatch.ts
```diff
@@ -7,6 +7,34 @@ import * as XState from 'xstate';
 import * as XStateModel from 'xstate/lib/model';
 import * as XStateActions from 'xstate/lib/actions';
 
+// dont hate the player, hate the game
+// https://github.com/microsoft/vscode-loader/issues/33
```
Andarist, line 11: the issue should get fixed in the future on the Monaco side so we should be able to remove this beautiful hack, yay 🎉

src/sourceMachine.ts
```diff
@@ -171,13 +169,14 @@ export const makeSourceMachine = (params: {
...
       preserveActionOrder: true,
       context: {
         ...sourceModel.initialContext,
-        sourceRawContent: params.data?.text || null,
-        sourceID: params.data?.id || null,
-        sourceProvider: params.data ? 'registry' : null,
+        sourceRawContent: params.sourceRegistryData?.text || null,
+        sourceID: params.sourceRegistryData?.id || null,
+        sourceProvider: params.sourceRegistryData ? 'registry' : null,
+        sourceRegistryData: params.sourceRegistryData,
```
Andarist, line 179: this is the reason why I have removed `GetSourceFileSSR` - I have wanted to add this to the context right away so the `<VizHead />` could always consume this from here. But since that SSR-only query had a different "return type" it was creating a type mismatch and I didn't want to fight with this - so I've just reused the other query on the backend to achieve type symbiosis.

src/sourceMachine.ts
```diff
@@ -171,13 +169,14 @@ export const makeSourceMachine = (params: {
 
   return createMachine<typeof sourceModel>(
     {
-      initial: 'checking_if_on_legacy_url',
+      initial: 'checking_initial_data',
```
Andarist, line 172: this seemed like a priority condition - if we have data then we definitely don't need to check anything else

src/utils.ts
```diff
@@ -35,18 +35,17 @@ export function isInternalEvent(eventName: string) {
...
+  const [Provider, useContext] =
+    createRequiredContext<TInterpreter>(displayName);
 
-  const createUseSelector = <Data>(
-    selector: (state: TInterpreter['state']) => Data,
-  ) => () => {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    return useSelector(useContext(), selector);
-  };
+  const createUseSelector =
```
Andarist, line 43: Just prettier stuff here 🤷‍♂️

src/utils.ts
```diff
@@ -210,7 +209,6 @@ export const parseEmbedQuery = (query?: NextRouter['query']): ParsedEmbed => {
   const parsedEmbed = {
     mode: EmbedMode.Viz,
     panel: EmbedPanel.Code,
-    panelIndex: 0,
```
Andarist, line 213: This was a leftover - missed during refactors in the other PR.

src/pages/[[...sourceFileId]].tsx
```diff
@@ -0,0 +1,62 @@
...
+  // dynamic pages always have `req.params` available
+  const sourceFileIdParam = ctx.params!.sourceFileId;
+
+  if (!sourceFileIdParam) {
+    return {
+      props: {
+        sourceRegistryData: null,
+      },
+    };
+  }
```
Comment (mattpocock) [inline src/pages/[[...sourceFileId]].tsx:24]: This makes me a little sad that we can't handle this just by a URL - it means that all first loads of the site will have to go through SSR, even though they won't resolve anything. That's some lambda compute time wasted. But I understand why it's here.

mattpocock (approved):
Saving is working - looks fine to me!

src/pages/[[...sourceFileId]].tsx
Comment (Andarist) [reply src/pages/[[...sourceFileId]].tsx:24]: Yeah, that's right and it saddens me quite a bit as well. OTOH - if we'd be able to use the authenticated user here (which I hope we'll be able to do in the future) then we could SSR user-related parts of the UI, even for paths without an ID.

farskid (approved):

PR #248, opened 2021-09-02
Improved lazy loading of prettier by only requiring it when it's first used

mattpocock:
This PR ensures that prettier is now only loaded when it's actually used for formatting for the first time. It turns `prettier.format` into an async call, which the first time it's called loads prettier via CDN (with a 15 second timeout).

Since it's loaded via CDN, subsequent requests to the page are done to the disk cache.

If you like this, we could also do this with Elk.

### DIFF
--- a/src/prettier.ts
+++ b/src/prettier.ts
@@ -1,0 +1,73 @@
+const loadScript = (src: string) => {
+  return new Promise<void>((resolve, reject) => {
+    const scriptElement = document.createElement('script');
+    scriptElement.src = src;
+
+    const timeout = setTimeout(() => {
+      reject('Timeout');
+    }, 15000);
+
+    scriptElement.onload = () => {
+      resolve();
+      clearTimeout(timeout);
+    };
+    document.body.appendChild(scriptElement);
+  });
+};
+
+declare global {
+  export const prettier: typeof import('prettier');
+  export const prettierPlugins: (string | import('prettier').Plugin)[];
+
+  interface Window {
+    define: any;
+  }
+}
+
+const Prettier = () => {
+  let hasLoaded = false;
+
+  const loadPrettier = async () => {
+    if (hasLoaded) return;
+
+    /**
+     * Helps Monaco editor get over a too-sensitive error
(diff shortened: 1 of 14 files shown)


davidkpiano:
> If you like this, we could also do this with Elk.

Yes please!

davidkpiano (approved):

src/prettier.ts
```diff
@@ -0,0 +1,73 @@
...
+  export const prettier: typeof import('prettier');
+  export const prettierPlugins: (string | import('prettier').Plugin)[];
+
+  interface Window {
+    define: any;
+  }
+}
+
+const Prettier = () => {
+  let hasLoaded = false;
```
davidkpiano, line 28: Implicit state machine 🙈  (fine for now)

src/prettier.ts
```diff
@@ -0,0 +1,73 @@
...
+     * Helps Monaco editor get over a too-sensitive error
+     *
+     * https://stackoverflow.com/questions/55057425/can-only-have-one-anonymous-define-call-per-script-file
+     */
+    const define = window.define;
+    window.define = () => {};
+
+    await Promise.all([
+      loadScript('https://unpkg.com/prettier@2.3.2/standalone.js'),
+      loadScript('https://unpkg.com/prettier@2.3.2/parser-typescript.js'),
```
farskid, line 43:
```suggestion
      loadScript('https://unpkg.com/prettier@latest/standalone.js'),
      loadScript('https://unpkg.com/prettier@latest/parser-typescript.js'),
```

farskid (approved):

src/prettier.ts
mattpocock: Is this a good idea? I think it might be better to fix it to a specific version

src/prettier.ts
mattpocock: 😂  I figured this was easier than trying to wrangle `service.send` inside an async function

davidkpiano:
Is this good to go in?

PR #255, opened 2021-09-07
Embed preview

farskid:
This PR adds a preview for embedding the visualization.

image

image


src/EmbedPreview.tsx
```diff
@@ -0,0 +1,384 @@
...
+              position="absolute"
+              rounded="false"
+              top="0"
+              right="0"
+              transform="translateY(-110%)"
+              onClick={copyEmbedCode}
+            >
+              {isCopied ? 'Copied' : 'Copy'}
+            </Button>
+            <Textarea
```
davidkpiano, line 326: Nit/future idea: can we make it so that clicking the textarea highlights all of its text content?

src/EmbedPreview.tsx
```diff
@@ -0,0 +1,384 @@
...
+                </FormLabel>
+                <Switch defaultChecked={false} id="zoom" name="zoom" />
+              </FormControl>
+            </VStack>
+          </form>
+          <Box position="relative" width="100%">
+            <Button
+              size="xs"
+              position="absolute"
+              rounded="false"
```
davidkpiano, line 318:
```suggestion
              rounded="none"
```

`false` isn't an option

davidkpiano (approved):
Looks good to me!

Two future ideas:

1. When adjusting the different embed options, as a user, I'd expect the embed preview to be instant. I believe we can accomplish this by dynamically changing the query parameters in the `<iframe>` without doing a full reload?
2. For "panels" mode, we should have a "viz" panel so everything is full-screen and we don't lose the visualization.

farskid:
> Looks good to me!
> 
> Two future ideas:
> 
>     1. When adjusting the different embed options, as a user, I'd expect the embed preview to be instant. I believe we can accomplish this by dynamically changing the query parameters in the `<iframe>` without doing a full reload?
> 
>     2. For "panels" mode, we should have a "viz" panel so everything is full-screen and we don't lose the visualization.

Point 1 is totally possible and a good idea.
I'm not sure I follow point 2.

src/EmbedPreview.tsx
farskid: I can do it right now.

src/App.tsx
```diff
@@ -90,6 +90,21 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
...
 
+  // TODO: Subject to refactor into embedActor
+  useEffect(() => {
+    window.onmessage = async (message) => {
+      const { data } = message;
+      console.log(data);
+      switch (data.type) {
+        case 'EMBED_PARAMS_CHANGED':
+          let x = await router.replace(data.url, data.url);
+          console.log(x);
```
davidkpiano, line 101:
```suggestion
      switch (data.type) {
        case 'EMBED_PARAMS_CHANGED':
          await router.replace(data.url, data.url);
```

mattpocock:
https://www.loom.com/share/6db55bfdbbcb4c62bf7d34dc8be29221

.changeset/gorgeous-hotels-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate-viz-app': minor
+---
+
+Adds the embed preview frame that lets users to configure how they want the embedded visualizer to look like
```
mattpocock, line 5: Could you add an image to this changeset?

src/App.tsx
```diff
@@ -90,6 +90,21 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
     });
   }, [machine?.id, sendToSourceService]);
 
+  // TODO: Subject to refactor into embedActor
+  useEffect(() => {
```
mattpocock, line 94: nit: Could you move this into a custom hook? That'll give it a bit more context as to what it's about.

src/EmbedPreview.tsx
```diff
@@ -0,0 +1,532 @@
...
+      },
+    }),
+  );
+
+  useEffect(() => {
+    // service.onEvent((evt) => console.debug(evt.type));
+    service.onTransition((state, ctx) => {
+      console.debug(state.event, state.actions);
+    });
+  }, []);
```
mattpocock, line 248:
```suggestion
```

src/EmbedPreview.tsx
```diff
@@ -0,0 +1,532 @@
...
+
+  useEffect(() => {
+    const formRef = form.current;
+    sendPreviewEvent({ type: 'IFRAME_READY', iframe: iframe.current });
+    return () => {
+      formRef.reset();
+    };
+  }, []);
```
mattpocock, line 256: What does this do? Reminds me a bit of my legacy code talk 😂 

Could you add a comment? Or wrap in a custom hook?

mattpocock (approved):
Approved with nits

farskid:
Update: Now options are conditionally shown based on relevance.

image

src/EmbedPreview.tsx
farskid: Hard to wrap in a custom hook. it does 2 things:
1. On mount, signal that iframe ref is ready
2. Reset the form when leaving the embed preview (unmount)

src/CanvasContainer.tsx
```diff
@@ -366,11 +362,9 @@ const dragMachine = dragModel.createMachine(
 );
 
 const getCursorByState = (state: AnyState) =>
-  (
-    Object.values(state.meta).find(
-      (m) => !!(m as { cursor?: CSSProperties['cursor'] }).cursor,
-    ) as { cursor?: CSSProperties['cursor'] }
-  )?.cursor;
+  (Object.values(state.meta).find(
+    (m) => !!(m as { cursor?: CSSProperties['cursor'] }).cursor,
```
mattpocock, line 366:
```suggestion
    (m) => Boolean(m as { cursor?: CSSProperties['cursor'] }).cursor,
```

src/App.tsx
```diff
@@ -18,7 +18,7 @@ import { theme } from './theme';
 import { EditorThemeProvider } from './themeContext';
 import { EmbedContext, EmbedMode } from './types';
 import { useInterpretCanvas } from './useInterpretCanvas';
-import { useRouter } from 'next/router';
+import router, { useRouter } from 'next/router';
```
mattpocock, line 21: Is this import used anywhere?

farskid: Yes. https://github.com/statelyai/xstate-viz/pull/255/files/d688d5f429313320ede52ce00c7bf105066d073c#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R101

mattpocock: Shouldn't this come from the `useRouter` itself? Not sure what using this singleton gives us in terms of benefit

farskid: oh right. yeah this can come from the named exports too.

Issue #250, opened 2021-09-02
"Send" button is disabled when trying to send an event with custom data


a-uz:
I already noticed two other issues (#222, #223) regarding the send button, but this issue seems different.

As mentioned in the title, "Send" button is disabled when trying to send an event with custom data.

![image](https://user-images.githubusercontent.com/6834140/131854263-69826431-c2f3-46c4-88a0-c236f7b4ebcb.png)

To reproduce this you may use this machine:

```ts
import { createModel } from "xstate/lib/model";

const counterModel = createModel(
  {
    count: 1,
  },
  {
    events: {
      INCREMENT: (by: number) => ({ by }),
    },
  }
);

const counterMachine = counterModel.createMachine({
  id: "counter",
  context: counterModel.initialContext,
  on: {
    INCREMENT: {
      actions: counterModel.assign({
        count: (context, event) => context.count + event.by,
      }),
    },
  },
});
```

With this event:

```json
{
  "type": "INCREMENT",
  "by": 1
}
```

Issue #223, opened 2021-08-27
"Send event" not working correctly in connected mode


Steve-OH:
When the visualizer is connected to another process, the "Send event" functionality doesn't work. The popup allows you to select a message to send, and the Send button _seems_ to be working, but nothing happens. No errors or other messages are displayed in the console.

)In standalone mode, the functionality works properly.)

Issue #222, opened 2021-08-27
"Send event" button enablement not working properly


Steve-OH:
The first time the "Send event" popup on the Events tab is displayed, the Send button is disabled, as expected, since there is no event to send. However, if you Cancel the popup, then open it again, this time the Send button is _not_ disabled.

PR #256, opened 2021-09-08
Fix a few bugs related to custom events

farskid:
Fixes https://github.com/statelyai/xstate-viz/issues/250 https://github.com/statelyai/xstate-viz/issues/222 https://github.com/statelyai/xstate-viz/issues/223

### DIFF
--- a/src/simulationMachine.tsx
+++ b/src/simulationMachine.tsx
@@ -303,7 +303,10 @@ export const simulationMachine = simModel.createMachine(
             (ctx, e) => {
               return {
                 type: 'xstate.event',
-                event: e.event,
+                event: {
+                  ...e.event,
+                  origin: e.event.origin || ctx.currentSessionId,
+                },
                 sessionId: ctx.currentSessionId,
               };
             },


src/EventsPanel.tsx
```diff
@@ -343,9 +343,13 @@ const newEventMachine = newEventModel.createMachine({
       on: {
         '*': [
```
Andarist, line 344: It's weird that we accept **any** event here but you assert that the event is of type `EVENT.PAYLOAD`. In such a case, shouldn't this be changed to this?
```suggestion
        'EVENT.PAYLOAD': [
```

Andarist: However, by looking at the code - this wouldn't be exactly correct because other events should result in the invalid state. In such a case I would propose to do the whole try/catch thing in the if block within `cond` that would be responsible for that event to make this more clear. Something like:
```js
if (e.type === 'EVENT.PAYLOAD') {
  // do the try/catch thing
} else {
  return false
}
```

Andarist: Or you could do this:
```js
on: {
  'EVENT.PAYLOAD': [{ cond: () => { /* try/catch thing */ }, target: '.valid' }, '.invalid'],
  '*': '.invalid'
}
```

src/simulationMachine.tsx
```diff
@@ -303,7 +303,10 @@ export const simulationMachine = simModel.createMachine(
             (ctx, e) => {
               return {
                 type: 'xstate.event',
-                event: e.event,
+                event: {
+                  ...e.event,
+                  origin: e.event.origin || ctx.currentSessionId,
+                },
```
Andarist, line 309: I don't think this is necessarily the correct thing to do. In my mind sending events from the viz resembles this piece of code:
```js
const service = interpret(machine.start)

onClick={() => service.send(createEvent())}
```
And in this scenario, the `origin` **wont** be set: https://codesandbox.io/s/hopeful-elion-0l4vg?file=/src/index.js

So, from my perspective, in this basic and most common scenario this wont resemble the production usage and thus it might introduce confusing behavior.

I recognize how not setting `origin` in some other scenarios might be "incorrect" as well though and it's something I've been mentioning to David some time ago. We've concluded though that, for the most part, this should be a less common scenario and that we won't bother with it at the moment. 

When this can be incorrect? When an event can only be received from another machine in the real application and when one uses the `origin` to do something (like responding to it). We can't currently determine possible sources though so we don't have a way to provide any hints about them (so they could be selected or something).

Keep in mind that sending events from the Viz to any non-root machine is often risky - in a sense that it might make the whole thing behave differently than in a production use case (unless your application literally is sending directly to such a machine, which is a valid use case). In a lot of scenarios, you could end up with some inconsistent state that way. One that would not be achievable by the real app.

Andarist (commented):
Could we also add some basic tests for this thing? 

It would also be nice if we could explain such changes, at least briefly. What has been previously assumed? What has happened? How it got fixed? And why the new approach is better? I had to figure this stuff out on my own when reviewing this  - and having some kind of an explanation/reasoning behind the changes would help me a lot.

src/EventsPanel.tsx
farskid: This is definitely subject to refactoring but they're not changed in the scope of this PR 🤷‍♂️. Let's fix in a separate PR?

src/simulationMachine.tsx
farskid: One of those scenarios in which not setting the origin is a breaking behavior is when we send a custom event to an external inspected source. How would we solve that without setting the origin to the current interpreted machine? Maybe conditionally doing it only if we're in the inspector mode?

farskid:
> Could we also add some basic tests for this thing?
> 
> It would also be nice if we could explain such changes, at least briefly. What has been previously assumed? What has happened? How it got fixed? And why the new approach is better? I had to figure this stuff out on my own when reviewing this - and having some kind of an explanation/reasoning behind the changes would help me a lot.

I'm on it.

src/simulationMachine.tsx
Andarist: Could we go through this case on a call on Monday? I would love to get a little bit more context about this and this would allow me to jump into this quicker

src/EventsPanel.tsx
Andarist: This free flow of my mind (😅 ) has been prompted by the fact that this transition is defined for `*` events but it casts the received event to `EVENT.PAYLOAD`, seemingly not caring about other event types that can actually be received here.

This has been introduced in this PR and was not a problem before (because before the event type was not used here). So I think this should be refactored somehow as part of this PR

src/simulationMachine.tsx
Andarist: I would probably recommend dropping the `toSCXMLEvent` conversion from here: https://github.com/statelyai/xstate-viz/blob/7cc57c836c111eb1695a76b02de8f3d689cf59a6/src/EventsPanel.tsx#L388

and doing that here:
```js
toSCXMLEvent(e.event, { origin: ctx.currentSessionId })
```

davidkpiano:
@farskid @Andarist Besides tests, what are the remaining changes here?

PR #261, opened 2021-09-09
Include sourcemaps in production

jasonLaster:
Source Maps will make it easier to inspect the visualizer with browser devtools and Replay.io


Andarist:
It's currently not working as expected - we are looking into this with Jason.

Andarist (approved):
It turns out that it works OK 😅

mattpocock (approved):

PR #263, opened 2021-09-10
Fixed an issue with right-side panels getting cut off after stretching them and shrinking back with the resize handle

Andarist:


src/EditorPanel.tsx
```diff
@@ -140,12 +140,13 @@ const editorPanelMachine = editorPanelModel.createMachine(
               src: async (ctx) => {
                 const monaco = ctx.monacoRef!;
                 const uri = monaco.Uri.parse(ctx.mainFile);
-                const getWorker = await monaco.languages.typescript.getTypeScriptWorker();
+                const getWorker =
```
Andarist, line 143: just prettier here

src/EditorPanel.tsx
```diff
@@ -436,7 +437,7 @@ export const EditorPanel: React.FC<{
         {simulationMode === 'visualizing' && (
           <>
             {/* This extra div acts as a placeholder that is supposed to stretch while EditorWithXStateImports lazy-loads (thanks to `1fr` on the grid) */}
-            <div style={{ minHeight: 0 }}>
+            <div style={{ minHeight: 0, minWidth: 0 }}>
```
Andarist, line 440: the actual fix

src/ResizableBox.tsx
```diff
@@ -100,6 +101,9 @@ const ResizeHandle: React.FC<{
       onPointerUp={() => {
         send(dragDropModel.events['DRAG.END']());
       }}
+      onPointerCancel={() => {
+        send(dragDropModel.events['DRAG.END']());
+      }}
```
Andarist, line 106: for some reason `cy.realSwipe` was finishing with a pointercancel event and not pointerup, unsure why that is but handling pointercancel here is a good idea either way

src/ResizableBox.tsx
```diff
@@ -120,9 +124,11 @@ export const ResizableBox: React.FC<ResizableBoxProps> = ({
     // into multiple lines
     <Box
       {...props}
-      {...(!disabled && {
-        width: `clamp(35rem, calc(35rem + ${widthDelta}px), 70vw)`,
-      })}
+      style={
+        !disabled
+          ? { width: `clamp(35rem, calc(35rem + ${widthDelta}px), 70vw)` }
+          : undefined
```
Andarist, line 130: this is a performance optimization - such dynamic values like this one should not be "pumped" through Emotion because each intermediate value will result in a separate class name being created and injected in the CSSOM. Using inline style for this is strictly better here

mattpocock:
Looks good. Got a before/after?

mattpocock (approved):

Andarist:
**Before**
Screenshot 2021-09-13 at 10 36 27
**After**
Screenshot 2021-09-13 at 10 36 35

src/ResizableBox.tsx
davidkpiano: That's weird. In real life, how would a pointercancel event occur?

davidkpiano (approved):

src/ResizableBox.tsx
Andarist: From the https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointercancel_event the most realistic ones for us are:

> A hardware event occurs that cancels the pointer activities. This may include, for example, the user switching applications using an application switcher interface or the "home" button on a mobile device.

and:

> The device's screen orientation is changed while the pointer is active.

I was able to trigger it only by selecting a text and dragging it outside of the current window. Funny enough - it didn't actually cancel the text dragging session, I was still able to drop it successfully onto the URL bar. 🤷‍♂️ 

No harm in handling this since it exists though - I agree though that I would prefer to know when this can happen in real life 😅

PR #271, opened 2021-09-16
Add changeset + tests for #214

davidkpiano:


.changeset/stupid-beers-clap.md
```diff
@@ -0,0 +1,11 @@
...
+'xstate-viz-app': patch
+---
+
+Multiple actors are now correctly visualized in the actors panel. This includes actors that are:
+
+- Created with `createMachine(...)` (automatically interpreted by default)
+- Invoked with `invoke: { ... }`
+- Spawned with `spawn(...)`
+
+<img width="935" alt="CleanShot 2021-09-16 at 09 05 59@2x" src="https://user-images.githubusercontent.com/1093738/133617666-e36d525b-37b2-49a7-9354-92b7d3cd62ac.png">
```
Andarist, line 11: this could use a better `alt`

cypress/integration/actors.spec.ts
```diff
@@ -0,0 +1,34 @@
+describe('Actors panel', () => {
+  it('should reset the list when the viz is resetted', () => {
```
davidkpiano, line 2:
```suggestion
  it('should show all spawned and invoked actors', () => {
```

src/ActorsPanel.tsx
```diff
@@ -112,6 +112,9 @@ export const ActorsPanel: React.FC = () => {
                   ? 'line-through'
                   : undefined
               }
+              title={`Actor: ${
```
Andarist, line 115: I see what you are doing here since Cypress won't find text spanning multiple elements by default. However, this IMHO is functionally not needed and in fact, might be a small annoying distraction to the user. It only contains a pretty much duplicated text that is already visually available. 

Rule of thumb -> don't add code to only facilitate tests (which sometimes can be broken, since it's just sometimes impossible)

Andarist (approved):

PR #276, opened 2021-09-22
Added changeset for OG image changes

mattpocock:


.changeset/selfish-maps-drive.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate-viz-app': minor
+---
+
+Added generated OG images to visualized machine links. You can copy these links yourself to embed an up-to-date image of your visualized machine - for instance in docs. Here's where the share button is in the viz:
+
+![](https://user-images.githubusercontent.com/28293365/134316316-70f2cd1d-aa05-409c-b058-4a46fc8af1b5.png)
```
davidkpiano, line 7:
```suggestion
![Share button on Stately Viz](https://user-images.githubusercontent.com/28293365/134316316-70f2cd1d-aa05-409c-b058-4a46fc8af1b5.png)
```

davidkpiano (approved):

.changeset/selfish-maps-drive.md
mattpocock: Not required, because it's explained by the text around it. Learned from @laurakalbag!

Andarist: I'm not entirely sure about that - VoiceOver on Mac shows this at the moment:
Screenshot 2021-09-23 at 09 42 15
which is rather not helpful for the user.

And even if the text mentions what might be on the image I'm uncertain if it's something that we can ever rely on since one doesn't have to go through the whole text before jumping to an image. Nor do they have to read a page in a top-down order - what if they go in the opposite direction and they land on an image?

.changeset/selfish-maps-drive.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate-viz-app': minor
+---
+
+Added generated OG images to visualized machine links. You can copy these links yourself to embed an up-to-date image of your visualized machine - for instance in docs. Here's where the share button is in the viz:
```
Andarist, line 5:
```suggestion

pr: #198

Added generated OG images to visualized machine links. You can copy these links yourself to embed an up-to-date image of your visualized machine - for instance in docs. Here's where the share button is in the viz:
```

.changeset/selfish-maps-drive.md
mattpocock: I guess there's no harm in duplication.

Andarist (approved):

Issue #191, opened 2021-08-16
Websocket Support (to be used as an Inspector)


parker-codes:
My use case is to run the Viz locally and connect it to machines in a [Tauri](https://tauri.studio/en/) WebView. Because I'm not using a browser, the message passing has to go through a websocket connection.

To sum up my vision of how I would do it:
1. Start up my app
2. Point inspector to `https://localhost:3315` or whatever I specify
3. Run inspector locally at desired port
4. Begin inspecting a machine

jacksteamdev:
This would be really awesome for debugging a machine during testing!

I'm currently converting my [complicated Rollup plugin](https://github.com/extend-chrome/rollup-plugin-chrome-extension) to use XState for orchestration, and had I some idea of being able to create a WebSocket during Jest setup and view the state of the plugin in the browser during testing. This is the basic idea:

```javascript
// jest.setup.js
import { inspect } from '@xstate/inspect/lib/server'
import { Server } from 'ws'

const server = new Server({ port: 8080, clientTracking: true })
inspect({ server })
// then navigate to https://statecharts.io/inspect?server=localhost:8080
```

I've tried with the old inspector at https://statecharts.io/inspect?server=localhost:8080, but no joy. The inspector does connect, I can see the WebSocket in the dev console Networks tab. It even receives a message: `{"type":"xstate.inspecting"}`. The tests are awaiting a never ending promise at the time of these screenshots, so it's not b/c execution is paused while at a breakpoint.

![image](https://user-images.githubusercontent.com/23390212/134552985-c13e6d1a-4546-4a62-879a-0f2232c23b7f.png)

The global var `__xstate__` is available in the debugger, and I can see the machines in `__xstate__.services`, but no messages like `service.register` are sent across the wire.

![image](https://user-images.githubusercontent.com/23390212/134552869-416cf446-db85-49e2-8147-854bdd5c762b.png)

Anyways, this would be an fun use case for WebSocket support. I'm really excited to see the ecosystem mature, I think being able to inspect the logic flow will be a valuable tool to help contributors.

parker-codes:
I just thought of another use case. It seems like I wouldn't even need to run a local version of the viz/inspector as long as the stately.ai/viz was sending data through a local (configurable) websocket connection. So the inspector could be anywhere, **including the VSCode Extension**!

Am I right?

PR #293, opened 2021-10-11
Add support for inspecting via websocket

jacksteamdev:
Based on discussion #191, this PR adds support for inspecting via websocket. Note the new nested states in `inspecting` called `websocket` and `window`.

New queries have been added to the query string that reflect the parameters for `createWebSocketReceiver`. `server` is required and `protocol` is optional (default is `"ws"`).

```url
http://localhost:3000/viz?inspect&server=localhost:8888
http://localhost:3000/viz?inspect&server=localhost:8888&protocol=ws
```

~~The event handlers in `inspecting` in the invoked `proxy` service are duplicated. Not sure how dry you want this, but we could pull them out into the module scope inside a function.~~ I've moved the inspect receiver into context to avoid duplicating receiver event handlers. Not sure if this is a problem, maybe someone can suggest a better strategy?

This functionality depends on a PR for `@xstate/inspect`: https://github.com/statelyai/xstate/pull/2728 Without that PR, this doesn't work on the server side (in my use case, at least).

image

### DIFF
--- a/.changeset/brave-shoes-flow.md
+++ b/.changeset/brave-shoes-flow.md
@@ -2,4 +2,4 @@
 'xstate-viz-app': minor
 ---
 
-Add support for inspecting via websocket
+You can now inspect via WebSocket. Add the WebSocket server url as a query parameter, for example `https://stately.ai/viz?inspect&server=ws://localhost:3000`
--- a/.changeset/brave-shoes-flow.md
+++ b/.changeset/brave-shoes-flow.md
@@ -1,0 +1,5 @@
+---
+'xstate-viz-app': minor
+---
+
+Add support for inspecting via websocket
--- a/.changeset/brave-shoes-flow.md
+++ b/.changeset/brave-shoes-flow.md
@@ -2,4 +2,4 @@
 'xstate-viz-app': minor
 ---
 
-You can now inspect via WebSocket. Add the WebSocket server url as a query parameter, for example `https://stately.ai/viz?inspect&server=ws://localhost:3000`
+You can now inspect via WebSocket. To do that you can add the WebSocket server url as a query parameter, for example `https://stately.ai/viz?inspect&server=ws://localhost:3000`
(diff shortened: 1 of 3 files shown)


davidkpiano:
Since this is an important PR, can you add some `it.todo(...)` tests to verify the functionality works as expected? I can help fill them in.

jacksteamdev:
@Andarist Any chance this could get a quick review?

src/simulationMachine.tsx
```diff
@@ -106,13 +110,26 @@ export const simulationMachine = simModel.createMachine(
...
+            const serverUrl = new URLSearchParams(window.location.search).get(
+              'server',
+            );
+
+            let receiver: InspectReceiver;
+            if (serverUrl) {
+              const [protocol, ...server] = serverUrl.split('://');
+              receiver = createWebSocketReceiver({
+                protocol: protocol as 'ws' | 'wss',
+                server: server.join('://'),
```
Andarist, line 122: This looks somewhat off and could lead to some invalid URL being passed here (or rather to an invalid URL being constructed [here](https://github.com/statelyai/xstate/blob/fd69727fe787196b69f83931ff6944f6e9bdd0ee/packages/xstate-inspect/src/browser.ts#L301)). We might slightly refactor this later but no need to block this PR on this nit-picky comment

.changeset/brave-shoes-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate-viz-app': minor
+---
+
+You can now inspect via WebSocket. Add the WebSocket server url as a query parameter, for example `https://stately.ai/viz?inspect&server=ws://localhost:3000`
```
Andarist, line 5:
```suggestion
You can now inspect via WebSocket. To do that you can add the WebSocket server url as a query parameter, for example `https://stately.ai/viz?inspect&server=ws://localhost:3000`
```

Issue #300, opened 2021-10-19
Actions don't execute if specified via withConfig


josmithua:
**Description**

Implementation details are not executed when provided with `withConfig` instead of second param of `createMachine`

**Expected Result**

I would expect the actions to execute

**Actual Result**

The actions are not executed

**Reproduction**

https://stately.ai/viz/ba523ff8-913d-4917-b9ba-1cdd7029f6ff

PR #301, opened 2021-10-19
Fixed an issue with machines created by `withConfig`/`withContext` not being captured

Andarist:
fixes https://github.com/statelyai/xstate-viz/issues/300

but... the console output after the fix is:
```
in-config log
withConfig log
in-config log
```
that is because we capture 2 machines here (and this is somewhat an expected result as per https://github.com/statelyai/xstate-viz/pull/214, just not really in this case). This shows the limits of the current eval-like approach. It's basically impossible to cover common scenarios correctly without static analysis, cc @davidkpiano

### DIFF
--- a/src/parseMachine.ts
+++ b/src/parseMachine.ts
@@ -6,18 +6,19 @@ import realmsShim from 'realms-shim';
 
 const realm = realmsShim.makeRootRealm();
 
-const wrapCallbackToPreventThis =
-  (callback: (...args: any[]) => void) =>
-  (...args: any[]) => {
+function wrapToPreventThis<F extends (...args: any[]) => any>(callback: F): F {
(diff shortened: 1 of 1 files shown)


davidkpiano:
> > The other solution here is to opt into reading from inspect(someMachine).start(), then interpreted machines can be visualized the same as how inspected machines are.
> 
> Yes, we could differentiate the behavior based on the `interpret` calls being in the source file but that's kinda a hidden feature, people wouldn't easily realize the difference on their own.
> 
> Either way - this probably ain't a discussion for now. While this is gonna be slightly surprising - it's better than not supporting `withConfig` at all, right? 🤔 cc @mattpocock

Yep, this solution works for now

PR #309, opened 2021-11-02
Add new source file integration

mattpocock:
Duplicate of #308, pulled out of dev to merge other things first

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -1,5 +1,5 @@
 schema:
-  - https://api-prod.stately.ai
+  - https://api-dev.stately.ai
 documents:
   - 'src/graphql/**/*.graphql'
 generates:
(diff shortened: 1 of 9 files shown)


farskid (approved):

codegen.yml
```diff
@@ -1,5 +1,5 @@
 schema:
```
Andarist, line 1: what those two files at the top of PR are for? they mentioned dev env now and I wonder how this works with multi-env deployements that we have, are those URLs overwritten anywhere per env? is it intentional for this to always point to dev?

mattpocock: This file is for generating the types for the GQL endpoint with graphql-codegen. Since the viz uses a very limited subset of the GQL schema, I think we're safe to point at the 'latest, freshest' version instead of only point to prod.

We can also specify this with an env variable at a later date, but for now I'm comfortable with pointing it at dev.

Andarist (approved):

PR #312, opened 2021-11-16
Upgrade XState to the latest version

Andarist:
Related to https://github.com/statelyai/xstate/discussions/2770#discussioncomment-1650122

### DIFF
--- a/.changeset/rare-singers-refuse.md
+++ b/.changeset/rare-singers-refuse.md
@@ -1,0 +1,5 @@
+---
+'xstate-viz-app': patch
+---
+
+Updated XState to its latest version ([4.26.0](https://github.com/statelyai/xstate/releases/tag/xstate%404.26.0)). Visualizer should be able to use the new goodies now, such as it should provide access to the `invoke.meta` object in the invoke creators.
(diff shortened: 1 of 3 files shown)


.changeset/rare-singers-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate-viz-app': patch
+---
+
+Updated XState to its latest version ([4.26.0](https://github.com/statelyai/xstate/releases/tag/xstate%404.26.0)). Visualizer should be able to use the new goodies now, such as it should provide access to the `invoke.meta` object in the invoke creators.
```
Andarist, line 5: Maybe we should remove the "such as" note? I'm not sure - I've kinda only included that because it prompted me to do this update but the release itself contains more than that. WDYT?

davidkpiano: It's fine as-is 👍

davidkpiano (approved):

farskid:
@Andarist Just for bookkeeping purposes, we forgot to change the version shown under the help button.
image

PR #314, opened 2021-11-17
Adjust hotkeys before the release

Andarist:


src/CanvasView.tsx
```diff
@@ -127,10 +127,10 @@ export const CanvasView: React.FC = () => {
               variant="secondary"
             />
             <IconButton
-              aria-label="Fit to view"
-              title="Fit to view"
+              aria-label="Fit to content"
+              title="Fit to content"
               icon={<CompressIcon />}
-              onClick={() => canvasService.send('FIT_TO_VIEW')}
+              onClick={() => canvasService.send('FIT_TO_CONTENT')}
```
farskid, line 133: why are we changing this term?

src/SettingsPanel.tsx
```diff
@@ -72,18 +69,10 @@ const KeyboardShortcuts = () => (
...
-              <Kbd>f</Kbd>
-            </VStack>
-          </Td>
-          <Td>Fit to view</Td>
-        </Tr>
-        <Tr>
-          <Td>
-            <VStack alignItems="flex-start">
-              <Kbd>r</Kbd>
+              <Kbd>Shift</Kbd> + <Kbd>1</Kbd>
```
farskid, line 72: Remove the `VStack` entirely in favor of:

image

src/CanvasView.tsx
Andarist: https://github.com/statelyai/registry/pull/196#discussion_r747213455

Andarist: This is the one that The Piano Man prefers

farskid (changes requested):
Pending the change requested in the comment

src/SettingsPanel.tsx
farskid: Because with `VStack` the `+` is not shown in the UI

farskid (approved):

PR #315, opened 2021-11-17
Tweak controls in embed

farskid:
This PR:
- Adds initial scale meta tag to scale viz on the entire viewport on mobile view
- Tweak canvas controls in embed mode
  - RESET and Fit To View are now available in all embed modes
  - Zoom in and out buttons are only available if controls and Zoom are both enabled
  - Hand tool (pan button) is only available if controls and pan are both enabled
  - Reset canvas button and Help button are no longer available in embed mode

### Default:
image

### Zoom enabled
image

### Pan enabled
image

### DIFF
--- a/.changeset/young-hats-deliver.md
+++ b/.changeset/young-hats-deliver.md
@@ -1,0 +1,10 @@
+---
+'xstate-viz-app': minor
+---
+
+Tweak controls options in embed mode
+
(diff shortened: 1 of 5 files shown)


laurakalbag (approved):
I have nothing much helpful to add to this except YAY

davidkpiano (approved):

PR #317, opened 2021-11-18
Fix wrong type acquisition URL

davidkpiano:
This PR fixes this issue:

CleanShot 2021-11-18 at 00 02 56@2x

Probably not the best fix, but it works.

### DIFF
--- a/src/typeAcquisition.ts
+++ b/src/typeAcquisition.ts
@@ -139,7 +139,10 @@ const mapModuleNameToModule = (name: string) => {
 };
 
 //** A really simple version of path.resolve */
-const mapRelativePath = (moduleDeclaration: string, currentPath: string) => {
+const mapRelativePath = (
+  moduleDeclaration: string,
+  currentPath: string,
+): string => {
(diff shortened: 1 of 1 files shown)


src/typeAcquisition.ts
```diff
@@ -569,7 +573,11 @@ const getDependenciesForModule = (
           throw `No outer module or path for a relative import: ${moduleToDownload}`;
         }
 
-        const absolutePathForModule = mapRelativePath(moduleToDownload, path);
+        let absolutePathForModule = mapRelativePath(moduleToDownload, path);
+
+        if (moduleToDownload === '.') {
+          absolutePathForModule += '/index';
+        }
```
Andarist, line 580:
```suggestion
        // patch the broken logic from the original `typeAcquisition`
        // TODO: replace the whole thing with `@typescript/ata` that has been published since our "fork" was created
        if (moduleToDownload === '.') {
          absolutePathForModule += '/index';
        }
```

Andarist (approved):

PR #323, opened 2021-11-29
Added an embed mode which works via the URL, not any data in our db

mattpocock:


src/App.tsx
```diff
@@ -120,6 +122,32 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
...
+  );
+
+  const showZoomButtonsInEmbed = useMemo(
+    () => !embed?.isEmbedded || (embed.controls && embed.zoom),
+    [embed],
+  );
+  const showPanButtonInEmbed = useMemo(
+    () => !embed?.isEmbedded || (embed.controls && embed.pan),
+    [embed],
+  );
```
Andarist, line 149: I would suspect that this can actually decrease performance than improve it ;p all of those are just booleans and computing them is trivial thus there is no need to memorize this stuff

mattpocock: @Andarist this is in draft! No reviews yet please

src/pages/view-only.tsx
```diff
@@ -0,0 +1,91 @@
...
+      isEmbedded: true,
+      originalUrl: withoutEmbedQueryParams(router.query),
+    }),
+    [router.query],
+  );
+  return (
+    <>
+      <Head>
+        <script src="https://unpkg.com/elkjs@0.7.1/lib/elk.bundled.js"></script>
+      </Head>
```
Andarist, line 57: won't this page need any OG tags? would it be hard to reuse the existing `AppHead` and just configure it to skip downloading Monaco and whatever that is not needed?

src/pages/view-only.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <EmbedProvider value={embed}>
+        <ChakraProvider theme={theme}>
+          <CanvasProvider value={canvasService}>
+            <SimulationProvider value={simulationService}>
+              <Box
+                data-testid="app"
+                data-viz-theme="dark"
+                as="main"
+                display="grid"
+                gridTemplateColumns="1fr auto"
```
Andarist, line 67: There is just one column here, right?
```suggestion
                gridTemplateColumns="1fr"
```

src/App.tsx
```diff
@@ -142,7 +170,16 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
...
                     <CanvasProvider value={canvasService}>
-                      <CanvasView />
+                      <CanvasView
+                        shouldEnableZoomOutButton={shouldEnableZoomOutButton}
+                        shouldEnableZoomInButton={shouldEnableZoomInButton}
+                        canShowWelcomeMessage={canShowWelcomeMessage}
+                        showControls={showControls}
+                        showZoomButtonsInEmbed={showZoomButtonsInEmbed}
+                        showPanButtonInEmbed={showPanButtonInEmbed}
+                        isEmbedded={embed?.isEmbedded}
```
Andarist, line 180: q: wasn't `CanvasView` able to just read this from the context like before? it is provided here, after all 🤔

mattpocock: My guy, stop reviewing this! It's in draft

src/pages/view-only.tsx
mattpocock: @Andarist Not sure it will need OG tags as it'll mostly be used for:

1. OG image generation
2. Embedding in PR comments

Maybe it might need it for the second one? Unsure.

Andarist: the second one might especially need them as whenever we share a URL somewhere publicly it can be copied and shared again

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { withReadyRouter } from '../withReadyRouter';
+import lzString from 'lz-string';
+
+const parseMachineFromQuery = (query: NextRouter['query']) => {
+  if (!query.machine) {
+    throw new Error();
+  }
+
+  if (Array.isArray(query.machine)) {
+    throw new Error();
```
Andarist, line 25:
```suggestion
    throw new Error("`machine` query param can't be an array");
```

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { theme } from '../theme';
+import { NextComponentWithMeta } from '../types';
+import { useInterpretCanvas } from '../useInterpretCanvas';
+import { parseEmbedQuery, withoutEmbedQueryParams } from '../utils';
+import { withReadyRouter } from '../withReadyRouter';
+import lzString from 'lz-string';
+
+const parseMachineFromQuery = (query: NextRouter['query']) => {
+  if (!query.machine) {
+    throw new Error();
```
Andarist, line 21:
```suggestion
    throw new Error("`machine` query param is required");
```

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  if (Array.isArray(query.machine)) {
+    throw new Error();
+  }
+
+  const lzResult = lzString.decompressFromEncodedURIComponent(query.machine);
+
+  if (!lzResult) throw new Error();
```
Andarist, line 30:
```suggestion
  if (!lzResult) throw new Error("`machine` query param couldn't be decompressed");
```

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const machineConfig = JSON.parse(lzResult);
+
+  // Tests that the machine is valid
+  return createMachine(machineConfig);
```
Andarist, line 35: optionally you could extend the message with `err.message`
```suggestion
  try {
    return createMachine(machineConfig);
  } catch {
    throw new Error("decompressed `machine` couldn't be used to `createMachine`")
  }
```

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+ * You can also use the typical embed controls
+ */
+const ViewOnlyPage = withReadyRouter(() => {
+  const canvasService = useInterpretCanvas({
+    sourceID: null,
+  });
+  const simulationService = useInterpret(simulationMachine);
+  const router = useRouter();
+
+  useMachine(viewOnlyPageMachine, {
```
Andarist, line 93: q: was a machine the right tool for a job here? 😅

src/pages/view-only.tsx
```diff
@@ -0,0 +1,144 @@
...
+    () => ({
+      ...parseEmbedQuery(router.query),
+      isEmbedded: true,
+      originalUrl: withoutEmbedQueryParams(router.query),
+    }),
+    [router.query],
+  );
+
+  return (
+    <>
```
Andarist, line 110: nit: this Fragment is redundant

src/withReadyRouter.tsx
```diff
@@ -0,0 +1,25 @@
+import { useRouter } from 'next/router';
+import React, { useEffect, useState } from 'react';
+
+/**
+ * Ensures that Next's router is always ready (i.e. has
+ * query params loaded)
+ */
+export const withReadyRouter = (WrappedComponent: any) => {
+  WrappedComponent.displayName = `WithReadyRouter${WrappedComponent.displayName}`;
```
Andarist, line 9:
```suggestion
  WrappedComponent.displayName = `WithReadyRouter(${WrappedComponent.displayName || WrappedComponent.name})`;
```

However, this logic is still somewhat simplistic and won't cover all cases - it's fine for our purposes though.

Andarist (approved):
Since we have a test setup here in place, we could add a basic test for this feature

src/pages/view-only.tsx
mattpocock: Lol, I started with a machine and couldn't stop when it got simpler

PR #328, opened 2021-12-19
fixing #372: transitions makred as potentials even though ...

jbouecke:
they are blocked by 'in' definition in that machine

### DIFF
--- a/src/TransitionViz.tsx
+++ b/src/TransitionViz.tsx
@@ -98,12 +98,17 @@ export const TransitionViz: React.FC<{
     return null;
   }
 
+  // extra check if the transition might be blocked by the 'in' property...
+  const isBlocked = !!definition.in && !state.matches(definition.in);
+
   const isDisabled =
     delay?.delayType === 'DELAYED_INVALID' ||
     !state.nextEvents.includes(definition.eventType);
+
   const isPotential =
-    state.nextEvents.includes(edge.transition.eventType) &&
-    !!state.configuration.find((sn) => sn === edge.source);
+    state.nextEvents.includes(definition.eventType) &&
+    !!state.configuration.find((sn) => sn === edge.source) &&
+    !isBlocked;
 
   return (
     <button


src/TransitionViz.tsx
```diff
@@ -98,12 +98,17 @@ export const TransitionViz: React.FC<{
     return null;
   }
 
+  // extra check if the transition might be blocked by the 'in' property...
+  const isBlocked = !!definition.in && !state.matches(definition.in);
```
Andarist, line 102: How this is handled in XState's core suggests that this is not that simple:
https://github.com/statelyai/xstate/blob/89fcc6d47f1890f09752d73d086fdd158898c194/packages/core/src/StateNode.ts#L852-L863

jbouecke: I think you are right. My code only works if state path is given, and not if it is a state id (#...). I am working on a generic solution right now.

jbouecke: I added special handling for 'custom state ids'. I tested it with the (adapted) machine defined in #328 .

src/TransitionViz.tsx
```diff
@@ -98,12 +102,24 @@ export const TransitionViz: React.FC<{
...
   }
 
+  // extra check if the transition might be blocked by the 'in' property...
+  const isBlocked =
+    typeof definition.in === "string" &&  // exists
+    definition.in.length > 0 &&           // with non empty content
+    (
+      definition.in[0] === "#"            // is 'custom id' or path?
+        ? !machine || !state.matches(machine.getStateNodeById(definition.in.substring(1)).path.join(machine.delimiter))
+        : !state.matches(definition.in)
```
Andarist, line 112: I'm pretty sure that this branch is not completely correct - the code that handled this in the core is [this](https://github.com/statelyai/xstate/blob/89fcc6d47f1890f09752d73d086fdd158898c194/packages/core/src/StateNode.ts#L858-L862):
```js
            matchesState(
              toStateValue(stateIn, this.delimiter),
              path(this.path.slice(0, -2))(state.value)
            )
```
and this simple `state.matches(definition.in)` call can't satisfy this because this call is independent of the current state node, while the check in the core is resolved with that in mind.

A sideways note is that I think the checks here (and in the core) could be simplified slightly if we'd use `state.configuration`, the overall "algorithm" could look like this:
```js
state.configuration.includes(resolveStateNodeCorrectly(...args))
```

jbouecke: You were again right. Sorry for taking that long... The updated branch takes care about the 'search'. It was working because the exampe contained always the 'complet' path in the 'in' statement. Now it works also for deeply nested compound nodes with path descriptors that does not go up the compete tree (see example below).

```javaScript
import { createMachine } from "xstate";

createMachine({
  id: "example",
  initial: "TL",
  states: {
    TL: {
      type: "parallel",
      states: {
        S1: {
          states: {
            A: {},
            B: {},
          },
          on: {
            A: {
              target: "S1.A",
              in: "S2.B"
            },
            B: {
              target: "S1.B",
              in: "#S2B"
            }
          },
        },
        S2: {
          initial: "A",
          states: {
            A: {
              on: {
                B: "B",
              },
            },
            B: {
              id: "S2B",
              on: {
                A: "A",
              },
            },
          },
        },
      },
    }
  }
});
```

jbouecke:
I plan to do some code cosmetics and more testing

jbouecke:
I filed a [bug](https://github.com/statelyai/xstate/issues/2926) in xstate core.

PR #334, opened 2022-01-09
Remove compaction on root node. Fixes #279

### LINKED ISSUE #279: Multiple root events break Viz chart
**Description**
If a multiple root events are created, the Viz don't display chart.

The following code don't work:
`
  {
    on: {
      LOAD: {},
      UPDATE: {},
    },
    states: {
      something: {
      }
    },
  }
`

The following code work as espected:
`
  {
    on: {
      UPDATE: {},
    },
    states: {
      something: {
      }
    },
  }
`

**Expected Result**

The Viz would display two arrow on the top of the main machine box

**Actual Result**

The Viz don't display anything

**Reproduction**

I have create a machine to reproduce it:
https://stately.ai/viz/83677847-53ed-4ba2-87da-4f80b6654cab

davidkpiano:
This PR removes compaction on the root node, which was causing ELK to fail with self-transitions on the machine.

CleanShot 2022-01-09 at 06 31 09@2x

### DIFF
--- a/.changeset/stupid-cherries-talk.md
+++ b/.changeset/stupid-cherries-talk.md
@@ -1,0 +1,20 @@
+---
+'xstate-viz-app': patch
+---
+
+Self-transitions on the machine will no longer cause graph layout to fail:
+
+```js
+import { createMachine } from 'xstate';
+
+const machine = createMachine({
+  on: {
+    // These will now display as expected
+    LOAD: {},
+    UPDATE: {},
+  },
+  states: {
+    something: {},
+  },
+});
+```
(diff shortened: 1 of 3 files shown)


Andarist (commented):
Can u add a simple test that would just verify that the machine gets visualized correctly?

davidkpiano:
> Can u add a simple test that would just verify that the machine gets visualized correctly?

Done, please re-review

Andarist (approved):
Looks OK from mobile :) out of curiosity - what does compaction usually do? Was this misconfigured on our end or can this be classified as Elk’s bug?

davidkpiano:
> Looks OK from mobile :) out of curiosity - what does compaction usually do? Was this misconfigured on our end or can this be classified as Elk’s bug?

It's an ELK issue, I believe. The replies here suggest that it has to do with compaction: https://github.com/kieler/elkjs/issues/98

PR #336, opened 2022-01-19
Strip unecessary permissions from generated embed iframe html

### LINKED ISSUE #285: Restrict embed iframe's permissions by default
Today, if a user uses the **Embed** option for sharing their machine, the default iframe is overly permissive.  By default the following permissions given:

* Allow:
  *  `accelerometer`, `ambient-light-sensor`, `camera`, `encrypted-media`, `geolocation`, `gyroscope`, `hid`, `microphone`, `midi`, `payment`, `usb`, `vr`, `xr-spatial-tracking`.
* Sandbox:
  * `allow-forms`, `allow-modals`, `allow-popups`, `allow-presentation`, `allow-same-origin`, `allow-scripts`

After a little testing, the only requirements are `sandbox="allow-same-origin allow-scripts"`. For security everywhere, can we please narrow down the scope of these permissions to the bare minimum.

kmannislands:
As noted in the ticket, the feature policy on the iframe generated by the "Share > Embed" button are far too liberal.

This removes all entries from the feature policy as none seem necessary and restricts sandbox restrictions to a reasonable level.

Closes #285

### DIFF
--- a/.changeset/chatty-peas-bake.md
+++ b/.changeset/chatty-peas-bake.md
@@ -1,0 +1,5 @@
+---
+'xstate-viz-app': patch
+---
+
+Remove unecessary feature policy claims from generated embed iframes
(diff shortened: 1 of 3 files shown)


davidkpiano (approved):

Andarist (approved):
We should add a patch changeset to kick off a release with this change

PR #341, opened 2022-01-26
Refactored to use XState typegen

mattpocock:


src/dragSessionTracker.ts
```diff
@@ -149,27 +152,22 @@ export const dragSessionTracker = dragSessionModel.createMachine(
           return ctx.session;
         },
       }),
-      clearSessionData: assign({
+      clearSessionData: assign((ctx) => ({
```
Andarist, line 155: hm, this might be worth investigating - ideally this unused argument/function shouldn't be needed

Andarist: Btw. the event here is typed as `never` but that is incorrect - this action is an exit action used in the active state and this should be able to receive at least `DRAG_SESSION_STOPPED`

src/canvasMachine.typegen.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface Typegen0 {
+  '@@xstate/typegen': true;
+  eventsCausingActions: {
+    assignAfterZoomIn: 'ZOOM.IN';
+    resetPosition: 'POSITION.RESET';
+    setPositionAfterSourceChanged: 'SOURCE_CHANGED';
+    assignZoomAfterFitToContent: 'FIT_TO_CONTENT';
+    assignViewBoxAfterFitToContent: 'FIT_TO_CONTENT';
+    persistPositionToLocalStorage: string;
```
Andarist, line 11: this is definitely wrong, cause like we've discussed - this should always be a union of string literals (or a single string literal)

mattpocock: Yep, this would be different if I did it with the extension now

Issue #342, opened 2022-01-31
Dragging doesn't work on Firefox


lfjsa:
**Description**

The dragging functionality is broken in Firefox.

**Expected Result**

- clicking the hand icon enters the "drag" state, allowing the user to drag the visualisation
- clicking it again exits the "drag" state, the user can no longer drag the visualisation

**Actual Result**

- clicking the hand icon seemingly enters the "drag" state, but it throws an error on mousemove instead of dragging. The rest of the application behaves as expected.
  - Console error: "Uncaught TypeError: can't access property "pointerId", n is undefined"
- when the hand icon is clicked again, the application crashes completely
  - Rendered error: "Application error: a client-side exception has occurred (see the browser console for more information)."
  - Console error: "A client-side exception has occurred, see here for more info: https://nextjs.org/docs/messages/client-side-exception-occurred"

**Reproduction**

0. macOS 12.0.1, Firefox 97.0b9 (64-bit)
1. Navigate to an [empty visualiser](https://stately.ai/viz)
2. In the empty screen, click "See an example" to load an example machine
3. Visualise the example either by hitting `cmd+enter` or clicking the `🪄 Visualize` button
4. Once loaded, click the hand icon in the bottom left of the visualiser
5. Try to drag, check console for errors
6. Click hand icon again
7. Stately crashes

**Additional context**

Tested it in Safari and Edge on macOS. Works in both.

Andarist:
I've just merged a fix for this: https://github.com/statelyai/xstate-viz/pull/343 . We should cut a new release soon(ish).

PR #343, opened 2022-02-01
Fixed panning/dragging interations (the ones that include pressing pointer down) in FireFox

Andarist:
https://github.com/statelyai/xstate-viz/issues/342

FF is using 0 as `pointerId` for the primary pointer so naturally this didn't pass the truthiness check here. Other browsers worked as they are using 1.


Andarist:
Yep, my hard-earned experience allowed me to fix this quickly :trollface:

PR #347, opened 2022-02-24
Fix simulating events on child actors. #326

### LINKED ISSUE #326: running parent machine with 2 children through invoke duplicates first child displayed and the one invoked by parent is not interactive
**Description**
Creating 1 parent machine and 2 children is making the viz create the 3 machines independently (without parent) first and then creating a new instance of the first child called from the parent but with this last one not being interactive. The viz shows the first child instance (without parent) but when viewing the invoked one (via Actors tab) such state chart cannot be interacted with.

**Expected Result**
When creating a child machine and then invoking it from the parent, the same child machine instance should be used and it should be possible to interact with it.

**Actual Result**
When creating a child machine and then invoking it from the parent, a new child machine instance is created and it's not possible to interact with it.

**Reproduction**

Given 2 children machines(authNavigationStateMachine and appNavStateMachine) and one parent appRuntimeStateMachine that invokes authNavigationStateMachine on init, the "Actors" tab shows 4, with the authNavigationStateMachine apparently duplicated:

-> authNavigationStateMachine (x:57)
-> appNavStateMachine (x:58)
-> appRuntimeStateMachine (x:59)
-> appRuntimeStateMachine (x:59) → authNavigationStateMachine (x:61)

the viz shows authNavigationStateMachine (x:57) which is not the one invoked by appRuntimeStateMachine (x:59) so when clicking the event that transitions to the authNavigationStateMachine final state it doesn't get back to the parent.

Going to the "Actors" tab and clicking on "-> appRuntimeStateMachine (x:59) → authNavigationStateMachine (x:61)" shows the authNavigationStateMachine but clicking on any event doesn't work.

**Additional context**

Link to the viz file.

https://stately.ai/viz/dcd508c8-0915-4c3f-9ad3-d881a1f37d0f#

MarianPalkus:
This PR fixes #326 by searching for child actors when simulating events:

- adds method `findChildService` to `src/utils.ts` to recursively search for child service with the target `sessionId`
- start the search for child services if no root service matches the `sessionId` of the even to process

### DIFF
--- a/src/simulationMachine.tsx
+++ b/src/simulationMachine.tsx
@@ -14,6 +14,7 @@ import { devTools } from './devInterface';
 import { notifMachine } from './notificationMachine';
 import { AnyState, AnyStateMachine, ServiceData } from './types';
 import { isOnClientSide } from './isOnClientSide';
+import { findChildService } from './utils';
 
 export interface SimEvent extends SCXML.Event<any> {
   timestamp: number;
@@ -180,7 +181,7 @@ export const simulationMachine = simModel.createMachine(
                   }
                 });
               } else if (event.type === 'xstate.event') {
-                const service = serviceMap.get(event.sessionId);
+                const service = serviceMap.get(event.sessionId) || findChildService(Array.from(serviceMap.values()), event.sessionId);
                 if (service) {
                   try {
                     service.send(event.event);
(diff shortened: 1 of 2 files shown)


MarianPalkus:
Might be related to #352.

MarianPalkus:
Sure, added a test case. I hope the it makes sense.

MarianPalkus:
Might be related to #348.

PR #367, opened 2022-06-21
Andarist/webext groundwork refactor

Andarist:
Goals of this PR:
- make the "layout" of the main app more customizable, or rather - achieve bigger control over the pieces that are rendered without a lot of conditional rendering tightly couple to context etc
- this allows us to render (in the future) `<WebExtension/>` that eg. doesn't depend on Monocao anyhow or doesn't know anything about embed mode etc
- remove dependency on SCSS to make it easier to bundle `<WebExtension/>` and also to migrate closer to a single source of truth for styling (previously we were mixing SCSS with Chakra)

In general, to review this PR it might be the easiest to just follow its commits. Most are self-contained refactoring pieces, with some minor bug fixes at the end.

I'm not super fond of some of those "customizable" points here - perhaps this could be structured in a better way. I feel though that the current shape of the code is manageable and we could always refactor stuff later. If you have any general thoughts on composition patterns for problems like this - I'm happy to discuss different tradeoffs etc.

### DIFF
--- a/cypress/integration/embedded-mode.spec.ts
+++ b/cypress/integration/embedded-mode.spec.ts
@@ -34,18 +34,22 @@ describe('Embedded mode', () => {
     });
   });
   describe('default (mode:viz)1', () => {
-    before(() => {
+    it('panels should be hidden', () => {
       cy.interceptGraphQL({
         getSourceFile: sourceFileFixture,
       });
       cy.visitEmbedWithNextPageProps({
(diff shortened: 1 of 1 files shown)


mellson (approved):
I think this restructuring makes sense. And from what I can test at the preview deploy, it also looks to be working. And the only failing test is fixed upstream. So LGTM 🚢

PR #382, opened 2022-08-22
The editor will now prefer imports from xstate core

mellson:
The autocomplete test for importing assign from xstate has been updated to support picking the new placement of assign in the autocomplete menu. The assign we want is now number three in the list, was number two before.

Update: hmm 🤔 it looks like our Cypress test suite here is unreliable. It's now failed three times, for three different reasons. This isolated change should be fine, but I guess we need another issue around getting these tests more stable.

### DIFF
--- a/cypress/integration/autocomplete.spec.ts
+++ b/cypress/integration/autocomplete.spec.ts
@@ -1,15 +1,3 @@
-const importFromXstate = () => {
-  cy.get('body').then(($body) => {
-    if ($body.text().includes("Auto import from 'xstate'")) {
-      cy.getMonacoEditor().type('{enter}');
-    } else {
-      cy.getMonacoEditor().type('{downarrow}');
-      cy.wait(500);
-      importFromXstate();
-    }
-  });
-};
-
 describe('Autocomplete in the editor', () => {
   it('Should allow you to autocomplete XState imports', () => {
     cy.visit('/viz');
@@ -18,7 +6,7 @@ describe('Autocomplete in the editor', () => {
     // Wait for the autocomplete to show up
     cy.contains('assign');
 
-    importFromXstate();
+    cy.getMonacoEditor().type('{downarrow}{enter}');
 
     cy.contains(`import { assign, createMachine } from 'xstate';`);
   });
--- a/cypress/integration/autocomplete.spec.ts
+++ b/cypress/integration/autocomplete.spec.ts
@@ -6,7 +6,7 @@ describe('Autocomplete in the editor', () => {
     // Wait for the autocomplete to show up
     cy.contains('assign');
 
-    cy.getMonacoEditor().type('{downarrow}{downarrow}{enter}');
+    cy.getMonacoEditor().type('{downarrow}{downarrow}{downarrow}{enter}');
 
     cy.contains(`import { assign, createMachine } from 'xstate';`);
   });
--- a/cypress/integration/autocomplete.spec.ts
+++ b/cypress/integration/autocomplete.spec.ts
@@ -1,3 +1,15 @@
+const importFromXstate = () => {
+  cy.get('body').then(($body) => {
+    if ($body.text().includes("Auto import from 'xstate'")) {
+      cy.getMonacoEditor().type('{enter}');
+    } else {
+      cy.getMonacoEditor().type('{downarrow}');
+      cy.wait(500);
+      importFromXstate();
+    }
+  });
+};
+
 describe('Autocomplete in the editor', () => {
   it('Should allow you to autocomplete XState imports', () => {
     cy.visit('/viz');
@@ -6,7 +18,7 @@ describe('Autocomplete in the editor', () => {
     // Wait for the autocomplete to show up
     cy.contains('assign');
 
-    cy.getMonacoEditor().type('{downarrow}{downarrow}{downarrow}{enter}');
+    importFromXstate();
 
     cy.contains(`import { assign, createMachine } from 'xstate';`);
   });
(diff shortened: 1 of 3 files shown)


farskid:
@mellson Can we query the page to find the option containing `assign` keyword from the Monaco suggestion box and not rely on the order?

mellson:
> @mellson Can we query the page to find the option containing `assign` keyword from the Monaco suggestion box and not rely on the order?

That's actually what we do now. But the problem is that there's no visible information other than the word assign. And we need the third option.

farskid:
> > @mellson Can we query the page to find the option containing `assign` keyword from the Monaco suggestion box and not rely on the order?
> 
> That's actually what we do now. But the problem is that there's no visible information other than the word assign. And assign appears three times; we need the third option.

I wonder why three options appear for assign and why all are marked as auto import?
image

mellson:
> > > @mellson Can we query the page to find the option containing `assign` keyword from the Monaco suggestion box and not rely on the order?
> 
> > 
> 
> > That's actually what we do now. But the problem is that there's no visible information other than the word assign. And assign appears three times; we need the third option.
> 
> 
> 
> I wonder why three options appear for assign and why all are marked as auto import?
> 
> image
> 
> 

They are from the different exports we provide. And only one of them from core ‘xstate’ (Monaco doesn't reflect this before the import is done). Which is the one we're looking for.

Andarist:
If all of those auto-suggestions are related to auto-imports but for different modules then perhaps we could filter those suggestions altogether and only leave the one that we want there?

farskid:
> If all of those auto-suggestions are related to auto-imports but for different modules then perhaps we could filter those suggestions altogether and only leave the one that we want there?

Yeah, either that or structure our declared types better so that TS can reflect detailed lib name in auto-suggestion.

mellson:
> If all of those auto-suggestions are related to auto-imports but for different modules then perhaps we could filter those suggestions altogether and only leave the one that we want there?

I have tried to update the Monaco patch to include the code actions description when deciding on the auto import helper text.

And also created a small import helper in the autocomplete test to keep looking for the right one, so it's not hardcoded.

I'm not sure it's the right approach 🤔 I'm not even sure we should have a test like this. Are we testing the Monaco editor, or is this okay because we're trying to test that our own types show up?

Demo of how the test helper keeps going through the list until it finds the right import:
CleanShot 2022-08-22 at 16 34 57

src/monacoPatch.ts
```diff
@@ -123,12 +123,26 @@ function getAutoImport(provider: any, details: any): AutoImport | undefined {
...
-  } else if (details.name in XStateModel) {
-    specifier = 'xstate/lib/model';
-  } else if (details.name in XStateActions) {
+  if (
+    codeAction.description.includes('xstate/lib/actions') &&
+    details.name in XStateActions
+  ) {
     specifier = 'xstate/lib/actions';
+  } else if (
+    codeAction.description.includes('xstate/lib/actionTypes') &&
```
Andarist, line 132: any particular reason for this being added? I treat this rather as an internal module and I didn't expect it to be used externally

Andarist:
@mellson would it be hard to discard suggestions to add auto-imports to specifiers like `xstate/lib/actions` **if** the same import can be already found in `xstate`? The root entry point should always have priority here - and all those other specifiers will actually be removed in v5 so it would be nice to minimize their usage in the wild

src/monacoPatch.ts
mellson: the reason is that the current version of the viz shows `Auto import from ´xstate´` on three lines. But if you pick the first one it's actually an import from `xstate/lib/actionTypes`:

CleanShot 2022-08-22 at 16 45 24

mellson:
> @mellson would it be hard to discard suggestions to add auto-imports to specifiers like `xstate/lib/actions` **if** the same import can be already found in `xstate`? The root entry point should always have priority here - and all those other specifiers will actually be removed in v5 so it would be nice to minimize their usage in the wild

I don't know, it's my first time looking at the Monaco Editor today 😊 I'll try to see if it's possible to limit the number of providers in a smart way tomorrow.

mellson:
> Hm, it might be somewhat difficult if the completion list request is decoupled from the completion details request (and IIRC this actually might be the case). Most likely the required information to filter this out is in the response to this second request and it's only being performed "lazily" when you actually highlight a specific entry.

Not sure it's the right approach, but I've just pushed one possible solution. Where we check if an import exists in XState but is not from the right file. Looks to be working locally. What do you think?

farskid:
@mellson We might alternatively just filter suggestions:
image

image

mellson:
> @mellson We might alternatively just filter suggestions: image
> 
> image

The solution is filtering the list that becomes suggestions ☺️

cypress/integration/autocomplete.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('Autocomplete in the editor', () => {
     // Wait for the autocomplete to show up
     cy.contains('assign');
 
-    cy.getMonacoEditor().type('{downarrow}{downarrow}{enter}');
+    cy.getMonacoEditor().type('{downarrow}{enter}');
```
mellson, line 9: Now we only have one `assign` import in the suggestions, so one down arrow is enough.

farskid:
> > @mellson We might alternatively just filter suggestions: image
> > image
> 
> The solution is filtering the list that becomes suggestions ☺️

I thought that's the goal :)

farskid (approved):

PR #414, opened 2023-07-15
STA-4574 Add mixpanel analytics to xstate viz

kevinmaes:
CleanShot 2023-07-15 at 21 16 56@2x

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -113,7 +113,7 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
   }, [machine?.id, sendToSourceService]);
 
   useEffect(() => {
-    analytics()?.track('Opening XState Viz (localdev ignore)');
+    analytics()?.track('Opening XState Viz');
   }, []);
 
   // TODO: Subject to refactor into embedActor
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,6 +21,7 @@ import { useInterpretCanvas } from './useInterpretCanvas';
 import router, { useRouter } from 'next/router';
 import { parseEmbedQuery, withoutEmbedQueryParams } from './utils';
 import { registryLinks } from './registryLinks';
+import { analytics } from './analytics';
 
 const defaultHeadProps = {
   title: 'XState Visualizer',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -111,6 +112,10 @@ function App({ isEmbedded = false }: { isEmbedded?: boolean }) {
     });
   }, [machine?.id, sendToSourceService]);
 
+  useEffect(() => {
+    analytics()?.track('Opening XState Viz');
+  }, []);
```
Andarist, line 117: It's literally a thing that should happen once - I'd move this to the module scope. `useEffect` feels redundant here.

src/authMachine.ts
```diff
@@ -165,7 +166,10 @@ export const createAuthMachine = (params: {
               src: (_): Promise<LoggedInUser> =>
                 callAPI<LoggedInUser>({
                   endpoint: 'get-user',
-                }).then((res) => res.data),
+                }).then((res) => {
+                  analytics()?.identify(res.data.id);
```
Andarist, line 170: This one feels to me more like an action called in response to this async job - it's not exactly a core part of it. I'd move it to `onDone.actions`

src/App.tsx
Andarist: Are we fine with this being called this early? The `.identify` call might not happen yet at this point in time. But perhaps MixPanel is smart enough to correlate those 2 things based on their own user ID?

kevinmaes: We should verify the order here. Rather than call `track('Opening XState Viz');` in the module scope, we may need to defer the call.

src/authMachine.ts
kevinmaes: Good idea. Now that this is working, generally, I can move it to `onDone`.

src/App.tsx
kevinmaes: @Andarist You're right, the `identify()` call was being made after the initial tracking call in the App's `useEffect` which explains why the id wasn't picked up in the payload of the tracking call.

kevinmaes:
Unique ID is now being passed to the `analyze` api endpoint.
CleanShot 2023-07-16 at 09 16 21@2x

davidkpiano (approved):

